PR #133, opened 2015-01-01
Optionally generate docs for undocumented declarations

### LINKED ISSUE #129: Options for showing undocumented structures
There should be command-line flags for showing/hiding undocumented structures (separated by type like global function, global variable, selector/method, operator, class, structure, extension, ...). This way an API can be documented completely without having to write documentation to every structure (some are self-explanatory, but still name and type/signature are worthy information).

jpsim:
Closes #129. Please review @juangamnik and @segiddins.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,10 @@ None.
 * Combine abstract and discussion in page overview.  
   [JP Simard](https://github.com/jpsim)
   [#115](https://github.com/realm/jazzy/issues/115)
+* Optionally generate docs for all declarations, not just ones with
+  documentation comments.  
+  [JP Simard](https://github.com/jpsim)
+  [#129](https://github.com/realm/jazzy/issues/129)
 
 ##### Bug Fixes
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,17 +4,16 @@
 
 ##### Breaking
 
-None.
+* Don't skip declarations with no documentation comments by default.
+  Allow skipping using `--skip-undocumented`.  
+  [JP Simard](https://github.com/jpsim)
+  [#129](https://github.com/realm/jazzy/issues/129)
 
 ##### Enhancements
 
 * Combine abstract and discussion in page overview.  
   [JP Simard](https://github.com/jpsim)
   [#115](https://github.com/realm/jazzy/issues/115)
-* Optionally generate docs for all declarations, not just ones with
-  documentation comments.  
-  [JP Simard](https://github.com/jpsim)
-  [#129](https://github.com/realm/jazzy/issues/129)
 
 ##### Bug Fixes
 
--- a/lib/jazzy/config.rb
+++ b/lib/jazzy/config.rb
@@ -22,7 +22,7 @@ module Jazzy
     attr_accessor :root_url
     attr_accessor :version
     attr_accessor :min_acl
-    attr_accessor :document_all
+    attr_accessor :skip_undocumented
 
     def initialize
       self.output = Pathname('docs')
@@ -34,7 +34,7 @@ module Jazzy
       self.docset_platform = 'jazzy'
       self.version = '1.0'
       self.min_acl = SourceDeclaration::AccessControlLevel.internal
-      self.document_all = false
+      self.skip_undocumented = false
     end
 
     # rubocop:disable Metrics/MethodLength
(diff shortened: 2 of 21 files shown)


lib/jazzy/config.rb
```diff
@@ -33,6 +34,7 @@ def initialize
       self.docset_platform = 'jazzy'
       self.version = '1.0'
       self.min_acl = SourceDeclaration::AccessControlLevel.internal
+      self.document_all = false
```
segiddins: I feel like true should be the default

orta: +1 (it is in appledoc fwiw)

juangamnik: +1 should be true by default

jpsim: Alrighty then! `document_all` will be true, by popular demand!

jpsim:
Since "The People" have requested that declarations without any documentation comments be included in jazzy output by default, I've decided to offer a `--skip-undocumented` option rather than `--no-document-all`. It's cleaner and more descriptive, IMHO.

Could you take a l:eyes:k, @segiddins?

CHANGELOG.md
```diff
@@ -11,6 +11,10 @@ None.
 * Combine abstract and discussion in page overview.  
   [JP Simard](https://github.com/jpsim)
   [#115](https://github.com/realm/jazzy/issues/115)
+* Optionally generate docs for all declarations, not just ones with
```
segiddins: Please make a note of the default.

segiddins:
Why the change for the Alamofire spec?

segiddins:
Otherwise, :shipit:

jpsim:
> Why the change for the Alamofire spec?

Because @mattt seems to have explicitly left some declarations undocumented on purpose. I think a better approach would be to make those declarations private, however.

CHANGELOG.md
jpsim: Thanks. Updated.

segiddins:
Let's make a release with this?

-Samuel E. Giddins

On Jan 2, 2015, at 2:45 PM, JP Simard <notifications@github.com> wrote:

Merged #133.

mattt:
>  I think a better approach would be to make those declarations private, however.

The only purposefully undocumented parts of Alamofire are extensions to `NSURLRequest`, et al. to add conformance to the `URLRequestConvertible` and `URLStringConvertible` protocols. Any computed properties in those extensions are already documented by the protocol. I consider these to be implementation details; if this were Objective-C, these would be buried deep in a `.m` file somewhere.

jpsim:
@mattt so everything under `Alamofire.swift` in http://static.realm.io/jazzy_demo/Alamofire/undocumented.txt (except `public var URLString: String`) should be documented?

jpsim:
@segiddins released: https://github.com/realm/jazzy/releases/tag/v0.0.20

mattt:
@jpsim 

Going line by line:

### Alamofire.swift
	
        func queryComponents(key: String, _ value: AnyObject) -> [(String, String)]
	func escape(string: String) -> String

These two are internal functions, not exposed to API consumers.

	public var URLString: String
	public var URLString: String
	public var URLRequest: NSURLRequest

These are the aforementioned extensions.

	struct Singleton

This is an internal member, not exposed to API consumers.

	required public init(configuration: NSURLSessionConfiguration? = nil)

This is documented with a `:param:`. As an initializer, I thought a description to be redundant.

	deinit

This has since been removed. But as something that is both internal and not directly called, `deinit` should probably be special-cased as undocumentable.

	class SessionDelegate: NSObject, NSURLSessionDelegate, NSURLSessionTaskDelegate, NSURLSessionDataDelegate, NSURLSessionDownloadDelegate
	class TaskDelegate: NSObject, NSURLSessionTaskDelegate
	class DataTaskDelegate: TaskDelegate, NSURLSessionDataDelegate
	class UploadTaskDelegate: DataTaskDelegate
	class DownloadTaskDelegate: TaskDelegate, NSURLSessionDownloadDelegate
	func cURLRepresentation() -> String

These are all internal members. Again, not exposed to API consumers.

### NSURL.h

Not sure why these are included...

jpsim:
Thanks for clarifying, @mattt. Really, we should be using a `--min-acl` of `public` for libraries/frameworks.

The `NSURL.h` ones are showing up due to a bug in jazzy, which is turn due to an awkward behavior of SourceKit.

mattt:
Really, we should be using a --min-acl of public for libraries/frameworks.

That sounds great. Just having that command-line argument is really powerful and encouraging. Congrats on your great work so far. Really looking forward to seeing this become the de facto documentation tool for the emerging Swift community.

PR #143, opened 2015-01-06
[SourceKitten] De-duplicate declarations

segiddins:
De-duplicate the sidebar list of extensions and show all children for an extension, regardless of how many extensions on a type there are.  

@jpsim this might mess up sorting/ordering, so please double check that this is a change we want to make.


segiddins:
This was included in https://github.com/realm/jazzy/pull/148.

Issue #58, opened 2014-11-01
Extract default values from Podspec


jpsim:
Currently, many values have to be passed in to jazzy manually that could be automated. For example:

* author_name
* module_name
* github_url
* github_file_prefix
* author_url

Issue #66, opened 2014-11-01
Find a way to name extensions


jpsim:
Swift extensions aren't named/labeled, but it'd be nice to find a better way to present them:

![image](https://cloud.githubusercontent.com/assets/474794/4870672/2463abb0-617a-11e4-80a2-cf3810446a43.png)

Maybe by using the MARK-style comment directly above it?

khoogheem:
The other is with Extensions is not only are there multiple of the same name.. all of the links go to the same extension..  It builds all of the single files in the UIColor extension folder. but the uicolor.html is only showing what looks like the first extension of UIcolor it found

Doesn't Sourcekitten provide the filename in its output.. why not use that as part of the

jpsim:
@khoogheem actually @segiddins just fixed that! See https://github.com/realm/jazzy/blob/master/CHANGELOG.md#bug-fixes

khoogheem:
Ok.. I just saw that when using the master branch..

Here is my only problem now with Extensions )-

When I look at UIColor from the left tab bar it puts them all into one single long list of extensions.. (not sure if that is good or bad...)

Say I have NBA Colors in one file
NHL  Colors in another.
it lists them all  in the order of the way the file was read.

What is interesting is that i have //MARK: team name so I can better track it in Xcode.. which jazzy picks up and uses.

So you might get something like:
Atlanta Hawks
---hawsTeamcolors()
Chicago Bulls
-bullsTeamcolors()
Anaheim Ducks
-duckksTeamColors

If the file only has one //Mark:  it shows what you would expect
NBA Team Colors
-teamcolor()
-otherteamcolor()

NFL Team Colors
-teamcolor()
-otherteamcolor()

Would be nice if you could use the first Mark as the Main Header..

So your file might look like:
~~~
/**
	iOS & OSX Extensions of GKColor for Colors of NBA Teams

	**Swift Only**

	To get the red color of the Atlanta Hawks... AtlantaHawks.red.color()
*/
public extension GKColor {
    //MARK: NBA Colors

	//MARK: Atlanta Hawks
	/**
	Provides the Team Colors for the NBA Atlanta Hawks
	*/
	public enum AtlantaHawks {
		/** Color whose RGBA values are (r 0.0 g 0.0 b 0.502 a 1.0) */
		case blue
		/** Color whose RGBA values are (r 1.0 g 0.0 b 0.0 a 1.0) */
		case red
[...]

PR #148, opened 2015-01-13
Allow building documentation from only a Podspec

segiddins:
This already powers Swift documentation on CocoaDocs.
~~Depends upon https://github.com/jpsim/sourcekitten/pull/19.~~
Closes https://github.com/realm/jazzy/issues/58.

\c @orta @jpsim


jpsim:
https://github.com/jpsim/sourcekitten/pull/19 has been merged.

lib/jazzy/config.rb
```diff
@@ -35,6 +41,12 @@ def initialize
       self.version = '1.0'
       self.min_acl = SourceDeclaration::AccessControlLevel.internal
       self.skip_undocumented = false
+      self.podspec = nil
```
jpsim: we don't set any of the other attributes to `nil` since it's implicit.

lib/jazzy/config.rb
```diff
@@ -23,8 +24,13 @@ class Config
     attr_accessor :version
     attr_accessor :min_acl
     attr_accessor :skip_undocumented
+    attr_accessor :podspec
```
jpsim: Config options generally have an exposed command line option. Why not add them for `docset_icon`, `docset_path` and `source_directory`?

lib/jazzy/docset_builder.rb
```diff
@@ -8,20 +8,24 @@ class DocsetBuilder
...
       attr_reader :docset_dir
       attr_reader :documents_dir
 
       def initialize(output_dir, source_module)
-        @output_dir = output_dir + 'docsets'
         @source_module = source_module
-        @docset_dir = @output_dir + "#{source_module.name}.docset"
+        @docset_dir = output_dir +
+          (config.docset_path || "docsets/#{source_module.name}.docset")
+        @source_dir = output_dir
```
jpsim: this is confusing:

```ruby
@source_dir = output_dir
@output_dir = docset_dir.parent
```

Source is output, but output is docset's parent? Perhaps rename these variables to reflect the intention here?

lib/jazzy/podspec_documenter.rb
```diff
@@ -0,0 +1,109 @@
...
+    end
+
+    def sourcekitten_output
+      sandbox = Pod::Sandbox.new(pod_config.sandbox_root)
+      installer = Pod::Installer.new(sandbox, podfile)
+      installer.install!
+      stdout = Dir.chdir(sandbox.root) do
+        pod_targets.map do |t|
+          SourceKitten.run_sourcekitten(
+            %W(doc --module-name #{podspec.module_name} -target #{t}),
```
jpsim: should wrap `#{podspec.module_name}` in single quotes in case it contains a space

lib/jazzy/podspec_documenter.rb
```diff
@@ -0,0 +1,109 @@
...
+      when Pathname, String
+        require 'cocoapods'
+        podspec = Pod::Specification.from_file(podspec)
+      end
+
+      return unless podspec
+
+      config.author_name = author_name(podspec)
+      config.module_name = podspec.module_name
+      config.author_url = podspec.homepage || ''
```
jpsim: Won't this create empty links if `homepage` isn't set? Seems like there shouldn't be a default here.

jpsim: could default to `github_file_prefix(podspec)`

lib/jazzy/sourcekitten.rb
```diff
@@ -45,8 +45,11 @@ def self.make_doc_urls(docs, parents)
 
     # Run sourcekitten with given arguments and return STDOUT
     def self.run_sourcekitten(arguments)
-      bin_path = Pathname(__FILE__).parent + '../../bin'
-      `#{bin_path}/sourcekitten #{(arguments).join(' ')}`
+      bin_path = Pathname(__FILE__).parent + 'sourcekitten/sourcekitten'
```
jpsim: minor gripe: I'm now referring to the project as 'SourceKitten' but the command-line executable as 'sourcekitten', so I'd prefer if we changed to path to `SourceKitten/sourcekitten`

lib/jazzy/sourcekitten.rb
```diff
@@ -168,6 +171,14 @@ def self.doc_coverage
         (@undocumented_tokens.count + @documented_count)
     end
 
+    def self.deduplicate_declarations(declarations)
+      declarations.group_by { |d| [d.usr, d.type.kind] }.values.map do |decls|
```
jpsim: can you move `declarations.group_by { |d| [d.usr, d.type.kind] }.values` into a local variable? would help with legibility IMO.

jpsim:
Great work, Sam! The only thing I'm not convinced about is how extension deduplication is handled.

I actually find it preferable to name extensions (#66) and keep them individually paged rather than having all extensions of a given type on a single page.

I'm curious to see if @orta has feedback on this as well.

orta:
As an external lib user ( https://github.com/CocoaPods/cocoadocs.org/blob/master/cocoadocs.rb#L392-L418 ) it seems to be nice and logical. Considering a few of those options require me to reparse an appledoc template after doccing it's already saving a bunch of time. 

Won't really have a solid idea until I sit back and think about what I want it to look and feel like for CocoaDocs.

lib/jazzy/config.rb
segiddins: this is actually vestigial (used to serve a purpose) but will remove :+1:

lib/jazzy/config.rb
segiddins: sure

lib/jazzy/podspec_documenter.rb
segiddins: Modules must be C99-ext identifiers, so no space is allowed

lib/jazzy/podspec_documenter.rb
segiddins: :+1:

lib/jazzy/sourcekitten.rb
segiddins: Sure

lib/jazzy/sourcekitten.rb
segiddins: What would you call such a variable?

segiddins:
> I actually find it preferable to name extensions (#66) and keep them individually paged rather than having all extensions of a given type on a single page.

The only issue right now is that there is nothing to disambiguate extensions. Once you've added that, we can add that attribute to the `group_by` and it'll 'just work'. The de-duplication is not just for extensions, however. It's necessary for documenting podspecs, where the same file will be documented multiple times.

lib/jazzy/podspec_documenter.rb
jpsim: how convenient!

lib/jazzy/sourcekitten.rb
segiddins: Having trouble doing this because case-insensitive FS

lib/jazzy/config.rb
segiddins: done

lib/jazzy/sourcekitten.rb
jpsim: don't worry about it then. Like I said, minor gripe.

lib/jazzy/sourcekitten.rb
jpsim: That's kind of what I would have wanted to understand by reading a variable name :wink:

segiddins:
@jpsim I think I've addressed your comments

jpsim:
Ship it!

PR #151, opened 2015-01-14
Escape spaces in paths.

Baza207:
If a path is passed in with a space in it from the bin_dir or any arguments, these spaces are not escaped. This causes the path to be cut off at that point.

This probably requires a bit of testing and checking it doesn't need to be added anywhere else.

### DIFF
--- a/lib/jazzy/sourcekitten.rb
+++ b/lib/jazzy/sourcekitten.rb
@@ -1,5 +1,6 @@
 require 'json'
 require 'pathname'
+require 'shellwords'
 
 require 'jazzy/config'
 require 'jazzy/source_declaration'
@@ -45,7 +46,10 @@ module Jazzy
 
     # Run sourcekitten with given arguments and return STDOUT
     def self.run_sourcekitten(arguments)
-      bin_path = Pathname(__FILE__).parent + '../../bin'
+      bin_path = Shellwords.shellescape Pathname(__FILE__).parent + '../../bin'
+      arguments = arguments.map do |object|
+        Shellwords.shellescape object
+      end
       `#{bin_path}/sourcekitten #{(arguments).join(' ')}`
     end
 


Baza207:
Looks like the fact I've never coded in Ruby has made it fail Travis CI check for some simple issues. :(

jpsim:
Thanks for the PR, @Baza207. Although I think a more appropriate fix would be `Pathname(__FILE__).parent.parent.parent.join('bin')`.

Baza207:
I haven't coded in Ruby before, so you probably know best. This is what I managed to get working last night when I couldn't get it running. 😊

jpsim:
I'm not much of a rubyist either, but if you confirm that my suggestion works in your case, it seems a bit cleaner.

Baza207:
That works but I think Pathname(__FILE__).parent + '../../bin' feels like it describes what's happening more at a glance.

Though I didn't actually change that bit, as doing that doesn't help with the spaces not being escaped.

lib/jazzy/sourcekitten.rb
```diff
@@ -45,7 +46,10 @@ def self.make_doc_urls(docs, parents)
 
     # Run sourcekitten with given arguments and return STDOUT
     def self.run_sourcekitten(arguments)
-      bin_path = Pathname(__FILE__).parent + '../../bin'
+      bin_path = Shellwords.shellescape Pathname(__FILE__).parent + '../../bin'
```
segiddins: you shouldn't need to shell escape the `bin_path`

Baza207: Without it Jazzy always fails for me with `No such file or directory` error showing the `bin_path` only up to the first space.

segiddins: Really this just shows that we should be launching processes in a more proper way (I.e. popen3)

-Samuel E. Giddins

On Jan 15, 2015, at 1:54 AM, James Barrow <notifications@github.com> wrote:

In lib/jazzy/sourcekitten.rb:

> @@ -45,7 +46,10 @@ def self.make_doc_urls(docs, parents)
>  
>      # Run sourcekitten with given arguments and return STDOUT
>      def self.run_sourcekitten(arguments)
> -      bin_path = Pathname(__FILE__).parent + '../../bin'
> +      bin_path = Shellwords.shellescape Pathname(__FILE__).parent + '../../bin'
Without it Jazzy always fails for me with No such file or directory error showing the bin_path only up to the first space.

jpsim: Yes, using `popen3` sounds like a much safer fix to me as well.

Baza207: Is that something that would be added into config.rb either replacing OptionParser or working with it?

I'd love to be more help, but I think from here on things go out of my 2 day glance of Ruby knowledge. So I'm now more intrigued than much help I'm afraid.

jpsim:
This was fixed in #194. Thanks for your work here, @Baza207.

PR #152, opened 2015-01-14
Set the default encoding to UTF-8.

Baza207:
After making myself a little shell script to run on build of a scheme in Xcode, it would always fail on me, even though running it from the terminal worked fine. Setting the default encoding to UTF-8 seems to fix this issue.

Reference: https://github.com/berkshelf/berkshelf-api/issues/112 (near the bottom)
It's an encoding thing, notice how the JSON parser is trying to use US-ASCII as it's encoding. As of ruby 1.9 the encoding for JSON parser is explicit so you can either add this to the berks executable:
Encoding.default_external = "UTF-8"

### DIFF
--- a/lib/jazzy.rb
+++ b/lib/jazzy.rb
@@ -1,2 +1,4 @@
 require 'jazzy/config'
 require 'jazzy/doc_builder'
+
+Encoding.default_external = 'UTF-8'


Baza207:
Looks like the fact I've never coded in Ruby has made it fail Travis CI check for some simple issues. :(

segiddins:
You should just set the LANGE env var

Baza207:
Ok, I'm not sure how you would do that. Is thats something I'd do instead of the line I added or something that can be set when running jazzy?

trishayy:
I also had this problem running jazzy from Jenkins. I had to manually set the lang variable with the following shell command: export LANG=en_US.UTF-8

jpsim:
Thanks @Baza207!

Issue #159, opened 2015-01-30
Library not loaded: @rpath/libswiftCore.dylib sourcekitten


maltzsama:
I did uninstall of jazzy, updated gem, but nothing working
```
dyld: Library not loaded: @rpath/libswiftCore.dylib
  Referenced from: /usr/local/var/rbenv/versions/2.1.5/lib/ruby/gems/2.1.0/gems/jazzy-0.1.0/lib/jazzy/sourcekitten/sourcekitten
  Reason: image not found
/usr/local/var/rbenv/versions/2.1.5/lib/ruby/gems/2.1.0/gems/jazzy-0.1.0/lib/jazzy/sourcekitten.rb:51:in `run_sourcekitten': Running `/usr/local/var/rbenv/versions/2.1.5/lib/ruby/gems/2.1.0/gems/jazzy-0.1.0/lib/jazzy/sourcekitten/sourcekitten doc` failed:  (RuntimeError)
	from /usr/local/var/rbenv/versions/2.1.5/lib/ruby/gems/2.1.0/gems/jazzy-0.1.0/lib/jazzy/doc_builder.rb:55:in `build'
	from /usr/local/var/rbenv/versions/2.1.5/lib/ruby/gems/2.1.0/gems/jazzy-0.1.0/bin/jazzy:15:in `<top (required)>'
	from /usr/local/var/rbenv/versions/2.1.5/bin/jazzy:23:in `load'
	from /usr/local/var/rbenv/versions/2.1.5/bin/jazzy:23:in `<main>'
```

segiddins:
@maltzsama do you have Xcode (>= 6.1.1) installed as `/Applications/Xcode.app`?

ProjectDent:
I'm getting his same issue, and I have XCode 6.1.1 installed.

beltex:
Seems that for some reason, RubyGems is not downloading the symlink `Current -> A` inside of `lib/jazzy/sourcekitten/SourceKittenFramework.framework`, which is why the `rpath` is failing (as it expects `../Versions/Current/Frameworks`). This may not have been noticed before if you have installed SourceKitten standalone, as it installs the needed `dylib`s at `/Library/Frameworks/SourceKittenFramework.framework/Versions/Current/Frameworks` which is another valid `rpath` (can check via `otool -l` on the SourceKitten binary). `0.0.20` works fine, this is new since SourceKittens Makefile and install structure has changed a bunch.

For those who need a quick work around in the meantime.

```sh
[...]

Issue #144, opened 2015-01-07
Library not loaded error when running


Isuru-Nanayakkara:
After much trouble I finally installed Jazzy successfully. Then from the Terminal, I navigate inside an Swift Xcode project which I have documented according to the new [Swift documentation](http://nshipster.com/swift-documentation/) syntax and ran the command `jazzy`. But it outputs the following error.

**dyld: Library not loaded: @rpath/libclang.dylib
  Referenced from: /Library/Ruby/Gems/2.0.0/gems/jazzy-0.0.20/bin/sourcekitten
  Reason: image not found
Please pass in xcodebuild arguments using -x
If build arguments are correct, please file an issue on https://github.com/realm/jazzy/issues**

How do I resolve this?

beltex:
Where do you have Xcode installed? It's expected to be under `/Applications/Xcode.app/...`

You can check by running `xcode-select -p`

Isuru-Nanayakkara:
Yes. It returns this.
**/Applications/Xcode-Beta.app/Contents/Developer**

Its the correct path. I have the Xcode 6.2 beta installed.

beltex:
Not quite. Your's is `Xcode-Beta.app`, while expected is `Xcode.app`. I ran into this before [myself](https://github.com/beltex/SMCKit/issues/3#issuecomment-66921109). Simple work around is to create a symlink to your `lib` dir.

```bash
mkdir -p /Applications/Xcode.app/Contents/Developer/Toolchains/XcodeDefault.xctoolchain/usr
ln -s /Applications/Xcode-Beta.app/Contents/Developer/Toolchains/XcodeDefault.xctoolchain/usr/lib /Applications/Xcode.app/Contents/Developer/Toolchains/XcodeDefault.xctoolchain/usr/lib

Isuru-Nanayakkara:
I see. One question. When I eventualy install a stable Xcode version, do I have to remove the symlink or something?

beltex:
No. When you install a GM version either manually (by placing it in the `/Applications/` dir) or via the App Store it will overwrite that whole directory path and thus the symlink.

Is it working for you now?

Isuru-Nanayakkara:
Yep! That solved it. Works fine nwo. Thank you so much :+1:

beltex:
Nice, np! :)

PR #160, opened 2015-02-08
Better exception when Xcode not installed in standard location

AliSoftware:
This avoids SourceKitten to fail to launch and crash when Xcode is not installed in `/Applications/Xcode.app` (as SourceKitten would fail to load `libclang.dylib` in such cases)

See also #144 and [this comment](https://github.com/realm/jazzy/issues/159#issuecomment-73424745) on #159

### DIFF
--- a/lib/jazzy/sourcekitten.rb
+++ b/lib/jazzy/sourcekitten.rb
@@ -47,8 +47,7 @@ module Jazzy
     def self.run_sourcekitten(arguments)
       unless Pathname('/Applications/Xcode.app').exist?
         raise 'Xcode must be installed (or symlinked) at ' \
-        '/Applications/Xcode.app for jazzy to be able to work. ' \
-        'See https://github.com/realm/jazzy/issues/144 for more info.'
+        '/Applications/Xcode.app for jazzy to be able to work.'
       end
       bin_path = Pathname(__FILE__).parent + 'sourcekitten/sourcekitten'
       command = "#{bin_path} #{(arguments).join(' ')}"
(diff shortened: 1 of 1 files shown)


segiddins:
I don't like linking to GitHub issues in exception messages.

AliSoftware:
Ok, will remove it.

Can't access the Travis-CI website and report to check what went wrong there (Travis site seems to be down), any ideas about that?

AliSoftware:
Fixed.

lib/jazzy/sourcekitten.rb
```diff
@@ -45,6 +45,10 @@ def self.make_doc_urls(docs, parents)
 
     # Run sourcekitten with given arguments and return STDOUT
     def self.run_sourcekitten(arguments)
+      unless Pathname('/Applications/Xcode.app').exist?
+        raise 'Xcode must be installed (or symlinked) at ' \
+        '/Applications/Xcode.app for jazzy to be able to work.'
+      end
       bin_path = Pathname(__FILE__).parent + 'sourcekitten/sourcekitten'
       command = "#{bin_path} #{(arguments).join(' ')}"
```
AliSoftware: BTW I'm totally new to the `jazzy` codebase, but is it normal that there is no `shellescape` there (instead of `join(' ')`)?

segiddins: yes

jpsim:
Thanks @AliSoftware, we ended up doing this in #193.

PR #180, opened 2015-03-19
[SourceKitten] Allow using a :nodoc: token to skip specific tokens

segiddins:
When this is merged, I'll push up my addition to the integration specs

\c @jpsim


reitzig:
How does this work? It's not mentioned in the documentation, unfortunately.

jpsim:
Just include the string `:nodoc:` anywhere in the documentation comment for a declaration and it'll be skipped by jazzy. Here are a bunch of examples in Realm Swift: https://github.com/realm/realm-cocoa/search?utf8=%E2%9C%93&q=nodoc&type=

reitzig:
For reference: does not work for dropping `// MARK:`.

PR #184, opened 2015-03-23
[SourceKitten] No longer count initializers with parameters as undocumented

### LINKED ISSUE #183: Initializers with parameters are incorrectly detected as undocumented
This is because initializer parameters are detected by SourceKit as local variables and not parameters. Weird. This causes incorrect documentation coverage, as reported in #178.

jpsim:
Fixes #183.

Initializers with parameters would be counted additional times for each undocumented parameter (which SourceKit oddly categorizes as local variables instead of parameters). This meant that initializers showed up in `/undocumented.txt` and affected code coverage percentage.

Care to take a look @segiddins and @jessesquires?

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -13,7 +13,7 @@
 
 ##### Bug Fixes
 
-* No longer count initializers with parameters as undocumented.
+* No longer count initializers with parameters as undocumented.  
   [JP Simard](https://github.com/jpsim)
   [#183](https://github.com/realm/jazzy/issues/183)
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -13,7 +13,9 @@
 
 ##### Bug Fixes
 
-* None.
+* No longer count initializers with parameters as undocumented.
+  [JP Simard](https://github.com/jpsim)
+  [#183](https://github.com/realm/jazzy/issues/183)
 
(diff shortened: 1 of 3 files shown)


CHANGELOG.md
```diff
@@ -13,7 +13,9 @@
 
 ##### Bug Fixes
 
-* None.
+* No longer count initializers with parameters as undocumented.
```
segiddins: two spaces

lib/jazzy/source_declaration/type.rb
```diff
@@ -31,10 +31,19 @@ def mark?
...
+        declaration? && !param?
+      end
+
       def declaration?
         kind =~ /^source\.lang\.swift\.decl\..*/
       end
 
+      def param?
+        kind == 'source.lang.swift.decl.var.parameter' ||
+          kind == 'source.lang.swift.decl.var.local'
```
segiddins: please add a comment saying why `val.local` is a param

segiddins:
Do we need to add a test to `misc`?

jessesquires:
thanks @jpsim ! :+1:

jpsim:
All good comments, @segiddins. Thanks.

segiddins:
:+1:

PR #193, opened 2015-03-24
Raise exceptions if Xcode requirements aren't met.

jpsim:
/cc @segiddins

### DIFF
--- a/lib/jazzy/sourcekitten.rb
+++ b/lib/jazzy/sourcekitten.rb
@@ -55,12 +55,16 @@ module Jazzy
     def self.assert_xcode_location
       expected_xcode_select_path =
         Pathname('/Applications/Xcode.app/Contents/Developer')
-      actual_xcode_select_path = Pathname(`xcode-select -p`.chomp).realpath rescue nil
-      return if actual_xcode_select_path == expected_xcode_select_path
+      return if xcode_developer_directory == expected_xcode_select_path
       raise 'Please install Xcode 6.1 or 6.2 in ' \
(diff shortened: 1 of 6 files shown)


segiddins:
:+1: when green

lib/jazzy/sourcekitten.rb
```diff
@@ -52,8 +52,33 @@ def self.make_doc_urls(docs, parents)
       end
     end
 
+    def self.assert_xcode_location
+      expected_xcode_select_path =
+        Pathname('/Applications/Xcode.app/Contents/Developer')
+      return if xcode_developer_directory == expected_xcode_select_path
+      raise 'Please install Xcode 6.1 or 6.2 in ' \
```
AliSoftware: Especially as it is common to have multiple Xcode versions nowadays (with betas and all), and people don't always want to be forced to use a given strict name for their Xcode apps, I feel like creating a symlink should be also be an alternative listed in the exception message (if using a symlink still actually works obviously)?

segiddins: It does work

AliSoftware: Great :+1: this will probably be sufficient (no need for more, but at least it will be mentioned)
```diff
- raise 'Please install Xcode 6.1 or 6.2 in ' \
+ raise 'Please install or symlink Xcode 6.1 or 6.2 in ' \
```

jpsim: Done, thanks for the tip.

PR #202, opened 2015-04-03
Fix does not work on case sensitive file system

kishikawakatsumi:
Jazzy is not correctly installed when install on case sensitive file system. If run the Jazzy on case sensitive file system will cause an error such as the following:

```
$ bundle exec jazzy
dyld: Library not loaded: @rpath/libswiftCore.dylib
  Referenced from: /Users/katsumi/Desktop/jazzy-case-sensitive/vendor/bundle/ruby/2.2.0/gems/jazzy-0.1.4/lib/jazzy/sourcekitten/sourcekitten
  Reason: image not found
```

This error occurs due to Jazzy is not installed correctly. If you install Jazzy on case sensitive file system, `lib/jazzy/sourcekitten` directory is as follows:

There are both `sourcekitten` and `SourceKitten` directory, and symbolic link is also incorrect.

Correct is following:

Because the name of the actual directory and description of gemspec is different.

```
spec.extensions = ['lib/jazzy/sourcekitten/Rakefile']
```

```
jazzy/lib/jazzy/sourcekitten
```

So fix gemspec is the same as the name of the directory.

Could you please check it? @jpsim @segiddins


segiddins:
I'd rather just move the SourceKitten directory to have to proper case

kishikawakatsumi:
I see. I also think that is better.

Changed the SourceKitten directory name instead of changing the gemspec. And fixed execution command path of `sourcekitten` along with the change.

segiddins:
:+1: (the integration spec breakage is the CocoaPods update notice)

jpsim:
Thanks, @kishikawakatsumi!

PR #206, opened 2015-04-05
[Travis] Ignore new cocoapods version message

segiddins:
\c @jpsim

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -4,6 +4,7 @@ cache: bundler
 rvm:
   - system
 before_install:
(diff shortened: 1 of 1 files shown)


jpsim:
This will still fail locally. Can you pass a command line argument to have cocoapods not print this?

segiddins:
@jpsim not yet, but I've been thinking of making it an environment variable.

segiddins:
@jpsim this is passing now, good to :shipit: ?

jpsim:
Yes

PR #207, opened 2015-04-07
Updated SourceKitten to 0.3.2

jpsim:


spec/integration_spec.rb
```diff
@@ -88,7 +88,7 @@
       behaves_like cli_spec 'document_alamofire',
                             '-m Alamofire -a Alamofire ' \
                             '-u https://nshipster.com/alamofire ' \
-                            '-x -project,Alamofire.xcodeproj ' \
+                            '-x -project,Alamofire.xcodeproj,-dry-run ' \
```
segiddins: wait, I thought using `-dry-run` didn't work?

jpsim:
Clickable link to integration test fixtures: https://github.com/realm/jazzy-integration-specs/commit/dc1c3da2fbe1cc4ddc3f6b36fb2ecfe6b44e8657

spec/integration_spec.rb
jpsim: using `-dry-run` by default won't work for cases where there are non-apple dependencies, but that's not the case for either Alamofire or MiscJazzyFeatures.

segiddins:
@jpsim this is now green

PR #209, opened 2015-04-07
[SourceKitten] use SourceKitten's XML parsing

jpsim:
See changes to the integration specs here: https://github.com/realm/jazzy-integration-specs/compare/no-xml


lib/jazzy/sourcekitten.rb
```diff
@@ -120,41 +121,65 @@ def self.process_undocumented_token(doc, declaration)
...
+      declaration.return = make_paragraphs(doc, 'key.doc.result_discussion')
 
-      declaration.parameters = parameters_from_xml(xml)
+      declaration.parameters = parameters(doc)
 
       @documented_count += 1
     end
 
+    def self.string_until_first_rest_definition(string)
+      matches = /^\s*:[^\s]+:/.match(string)
```
segiddins: `/^\s*:[^\s:]+:/` ?

jpsim: no, `::` is a thing in ReST.

lib/jazzy/sourcekitten.rb
```diff
@@ -120,41 +121,65 @@ def self.process_undocumented_token(doc, declaration)
...
-              XMLHelper.xpath(parameter_el, 'Discussion') || '',
-            ),
-        }
+    def self.make_paragraphs(doc, key)
+      return nil unless doc[key]
+      doc[key].map do |p|
+        if p.keys.include?('Para')
+          Jazzy.markdown.render(p['Para'])
+        elsif p.keys.include?('Verbatim')
+          Jazzy.markdown.render("```\n" + p['Verbatim'] + "```\n")
```
segiddins: this loses the language for code highlighting?

segiddins: please use string interpolation

jpsim: yes, but we don't have that info at all from `Verbatim` tags anyway, plus if we're going to do our own ReST definition parsing, this is only temporary. This code path is only triggered for the contents of a ReST definition (practically `:param:` & `:returns:`). The comment body still goes straight through redcarpet.

lib/jazzy/sourcekitten.rb
```diff
@@ -120,41 +121,65 @@ def self.process_undocumented_token(doc, declaration)
...
-      declaration.abstract = XMLHelper.xpath(xml, 'Abstract')
-      declaration.discussion = XMLHelper.xpath(xml, 'Discussion')
-      declaration.return = XMLHelper.xpath(xml, 'ResultDiscussion')
-
-      nodoc = ->(string) { string.to_s.include? '<dt>nodoc</dt>' }
-      return if nodoc[declaration.abstract] || nodoc[declaration.discussion]
+      stripped_comment = string_until_first_rest_definition(
+        doc['key.doc.comment'],
+      ) || ''
+      declaration.abstract = Jazzy.markdown.render(stripped_comment)
```
segiddins: the abstract should only be the first line, with the discussion being the remainder

lib/jazzy/sourcekitten.rb
segiddins: :+1:

segiddins:
@jpsim this should be good to go!

jpsim:
@segiddins are the CSS changes in this PR correct?

segiddins:
Yes, they're what Joe and I worked on.

PR #219, opened 2015-04-30
[Config] Added ability to exclude files from documentation. Fixes #173.

### LINKED ISSUE #173: Allow ignoring files
Hi there,

is there a way to leave files out of the documentation ?
Like the appdelegate. I dont wanna document that one :D

And ofc, thanks for dat awesome tool

Regards

jpsim:
File path has to match exactly for this to work. Should it instead match on `end_with?()` or some other alternative like match a regular expression? /cc @bryanbartow @K4stor @segiddins 

For example, current logic:

```ruby
json.map do |doc|
  key = doc.keys.first
  doc[key] unless excluded_files.include?(key)
end.compact
```

potential alternative:

```ruby
json.map do |doc|
  key = doc.keys.first
  doc[key] unless
    excluded_files.select { |file| file.end_with?(key) }.count > 0
end.compact
```


lib/jazzy/config.rb
```diff
@@ -42,6 +43,7 @@ def initialize
       self.min_acl = SourceDeclaration::AccessControlLevel.public
       self.skip_undocumented = false
       self.source_directory = Pathname.pwd
+      self.excluded = []
```
segiddins: `excluded_files`

lib/jazzy/sourcekitten.rb
```diff
@@ -240,12 +240,19 @@ def self.deduplicate_declarations(declarations)
       end
     end
 
+    def self.filter_excluded_files(json)
+      excluded_files = Config.instance.excluded
+      json.map do |doc|
```
segiddins: why `map` instead of just a `reject` or `select`?

jpsim: Because I'm returning `doc[key]` and not `doc`.

segiddins: any reason, since that wasn't done before?

jpsim: This PR bumps SourceKitten to version 0.4.1, which includes the paths of the files being documented.

Format was previously `[{docs1}, {docs2}, ...]` and now it's `[{file1: {docs1}}, {file2: {docs2}}, ...]`.

PR #237, opened 2015-05-25
Added a 'Troubeshooting' section to README

Nickology:
Some issues seem to come up repeatedly and could maybe be avoided by having a small *troubleshooting* section in the README. 

As an example, I've added one question relating to the `--min-acl` parameter which is often the reason why users only see extensions when everything else seems to be running well.

### DIFF
--- a/README.md
+++ b/README.md
@@ -32,14 +32,10 @@ To install jazzy, run `[sudo] gem install jazzy` from your command line.
 Run `jazzy` from your command line. Run `jazzy -h` for a list of additional
 options.
 
-### Troubleshooting
+#### Troubleshooting
 
 ***Only extensions are listed in the documentation.***
-By default, jazzy will only create documentation for public methods and classes. If you want to generate documentation for all of your files, make sure you set the `--min-acl` flag to `internal` or `private`.
-
-```sh
-jazzy --output /path/to/output --source-directory /path/to/src --min-acl internal
-```
+By default, jazzy only documents public declarations. To generate documentation for declarations with a lower accessibility level (internal or private), please set the `--min-acl` flag to `internal` or `private`.
 
 ### Development
 
--- a/README.md
+++ b/README.md
@@ -32,6 +32,15 @@ To install jazzy, run `[sudo] gem install jazzy` from your command line.
 Run `jazzy` from your command line. Run `jazzy -h` for a list of additional
 options.
 
+### Troubleshooting
+
+***Only extensions are listed in the documentation.***
+By default, jazzy will only create documentation for public methods and classes. If you want to generate documentation for all of your files, make sure you set the `--min-acl` flag to `internal` or `private`.
+
+```sh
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -32,6 +32,15 @@ To install jazzy, run `[sudo] gem install jazzy` from your command line.
 Run `jazzy` from your command line. Run `jazzy -h` for a list of additional
 options.
 
+### Troubleshooting
+
+***Only extensions are listed in the documentation.***
```
jpsim: Please use an h4 header here.

README.md
```diff
@@ -32,6 +32,15 @@ To install jazzy, run `[sudo] gem install jazzy` from your command line.
...
+By default, jazzy will only create documentation for public methods and classes. If you want to generate documentation for all of your files, make sure you set the `--min-acl` flag to `internal` or `private`.
```
jpsim: Could you reword this a bit?

> By default, jazzy only documents public declarations. To generate documentation for declarations with a lower accessibility level (internal or private), please set the `--min-acl` flag to `internal` or `private`.

README.md
```diff
@@ -32,6 +32,15 @@ To install jazzy, run `[sudo] gem install jazzy` from your command line.
...
+
+```sh
+jazzy --output /path/to/output --source-directory /path/to/src --min-acl internal
```
jpsim: I think the paragraph above is sufficient to describe how to use the `--min-acl` parameter. I'd remove this sample usage snippet.

jpsim:
Thanks for the PR, @Nickology! I do think that this troubleshooting section could help new users understand how jazzy works. I've posted a few suggested edits that I'd like to hear your thoughts on before we can merge this.

However, I think jazzy should also print the number of internal and private declarations skipped to the console after it's done processing. This would probably address the problem at an earlier stage.

Nickology:
Hi @jpsim!

Your suggested edits are spot on. I'm going to make the changes and update the PR.
Totally agree with #238. My first reflex when faced with the blank documentation was to enable verbose mode which unfortunately did not provide any help in understanding the issue. Having more feedback on the console will definitely avoid confusion, especially when on-boarding.
This is a great project btw!

README.md
```diff
@@ -32,6 +32,11 @@ To install jazzy, run `[sudo] gem install jazzy` from your command line.
 Run `jazzy` from your command line. Run `jazzy -h` for a list of additional
 options.
 
+#### Troubleshooting
```
jpsim: Oops, I meant make the emphasized text an h4 header, not the "Troubleshooting" title! Like so:

```md
### Troubleshooting

#### Only extensions are listed in the documentation.

By default, jazzy only documents public declarations. To generate documentation for declarations with a lower accessibility level (internal or private), please set the `--min-acl` flag to `internal` or `private`.
```

jpsim:
Almost there, @Nickology!

Nickology:
Done :-)

jpsim:
:+1: many thanks

PR #245, opened 2015-05-30
Encrypt Travis Slack token

maclover7:


jpsim:
Thanks for the PR, but I'll pass on this because 1) encrypting the same token that was previously available in plaintext publicly offers limited protection and 2) when I run `travis encrypt` on the token, I get a different result.

PR #247, opened 2015-05-31
Updated the -h for excluded files

khoogheem:
option to include the requirement for exclude files needing absolute path


segiddins:
Honestly, I'd prefer to just allow relative paths for that?

khoogheem:
That is my feeling as well.  Right now it is absolute.  so I wanted to get at least the -h updated to alert people to it.. so they didn't have to read the Issue number that added in the ignore.

jpsim:
Can you instead, just make this work with relative paths?

Issue #20, opened 2014-07-03
Pass in arbitrary Mustache templates


jpsim:

jpsim:
Done in #257.

PR #257, opened 2015-07-07
[Templates] Added support for custom templates (addresses #20)

jpsim:
I also took advantage of this opportunity to move all bundled mustache templates into a single directory. /cc @segiddins 

I'll add documentation for this to the README in a minute.


segiddins:
:+1:

jpsim:
I would have preferred an approach where one could provide templates to *override* rather than requiring the entire template tree to be provided, even if only overriding a single part of it, but this won't stop me from merging this in.

PR #258, opened 2015-07-23
Set Dash width to 100%

Kapeli:
`width:auto` seems to cause issues like in the screenshot below. `width:100%` seems to work much better. Haven't tested it extensively though...

screen shot 2015-07-23 at 04 23 32


jpsim:
That looks much better! We'll want to test different configurations a bit before committing to this change though...

segiddins:
@jpsim this is just for dash, so I'm sure it's fine?

jpsim:
Yes, we should have just merged it originally. Needs an update to the integration specs though.

jpsim:
Thanks for the contribution, @Kapeli !

PR #262, opened 2015-07-29
make assets directory configurable enabling the use of own css / js

gurkendoktor:
Thanks for creating the possibility to pass a template directory. However, if you want to use different assets than the ones provided, you should also be able to be flexible with this. I created a parameter `--`-`--assets_directory`, that enables you to provide a different asset location. 

Please check backward compatibility and if everything is fine, you could merge this.

### DIFF
--- a/lib/jazzy/config.rb
+++ b/lib/jazzy/config.rb
@@ -31,6 +31,7 @@ module Jazzy
     attr_accessor :source_directory
     attr_accessor :excluded_files
     attr_accessor :template_directory
+    attr_accessor :assets_directory
 
     def initialize
       PodspecDocumenter.configure(self, Dir['*.podspec{,.json}'].first)
@@ -47,6 +48,7 @@ module Jazzy
       self.source_directory = Pathname.pwd
       self.excluded_files = []
       self.template_directory = Pathname(__FILE__).parent + 'templates'
+      self.assets_directory = Pathname(__FILE__).parent + '../../lib/jazzy/assets/'
     end
 
     def podspec=(podspec)
@@ -172,6 +174,11 @@ module Jazzy
         opt.on('t', '--template-directory DIRPATH', 'The directory that ' \
                'contains the mustache templates to use') do |template_directory|
           config.template_directory = Pathname(template_directory)
+        end
+
+        opt.on('--assets-directory DIRPATH', 'The directory that ' \
+               'contains the assets (CSS, JS, images)  to use') do |assets_directory|
+          config.assets_directory = Pathname(assets_directory)
         end
 
         opt.on('--readme FILEPATH',
--- a/lib/jazzy/config.rb
+++ b/lib/jazzy/config.rb
@@ -48,7 +48,7 @@ module Jazzy
       self.source_directory = Pathname.pwd
       self.excluded_files = []
       self.template_directory = Pathname(__FILE__).parent + 'templates'
-      self.assets_directory = Pathname(__FILE__).parent + '../../lib/jazzy/assets/'
+      self.assets_directory = Pathname(__FILE__).parent + 'assets'
     end
 
     def podspec=(podspec)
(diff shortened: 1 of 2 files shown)


lib/jazzy/config.rb
```diff
@@ -174,6 +176,11 @@ def self.parse!
           config.template_directory = Pathname(template_directory)
         end
 
+        opt.on('--assets-directory DIRPATH', 'The directory that ' \
+               'contains the assets (CSS, JS, images)  to use') do |assets_directory|
+          config.assets_directory = Pathname(assets_directory)
```
segiddins: can we throw a `.expand_path` at the end here?

lib/jazzy/doc_builder.rb
```diff
@@ -160,7 +160,8 @@ def self.write_undocumented_file(undocumented, output_dir)
     end
 
     def self.copy_assets(destination)
-      origin = Pathname(__FILE__).parent + '../../lib/jazzy/assets/.'
+      # origin = Pathname(__FILE__).parent + '../../lib/jazzy/assets/.'
```
segiddins: let's get rid of the commented-out code?

segiddins:
Looks great, thanks @gurkendoktor !

lib/jazzy/config.rb
gurkendoktor: sure, sorry for leaving this out

lib/jazzy/doc_builder.rb
gurkendoktor: i should be more thorough - sorry. can you remove this or should i amend it?

segiddins: you can either amend it or push a new commit removing it :)

jpsim:
I'd prefer moving the assets folder into the templates directory so it's all one thing.

gurkendoktor:
that could be also a way, then this one would be obsolete. i thought of this at first - however, this is kind of a major change in my opinion and I didn't want to break everything just by moving stuff around

jpsim:
Ok, let's stick with this approach, decoupling the assets directory with the templates! Makes sense.

By the way, CI is failing because the assets aren't being copied at the moment, so the integration tests don't match up. https://travis-ci.org/realm/jazzy/builds/73231977

gurkendoktor:
thanks for letting me know, i'll try to fix this asap. looks like the default config is not being used

lib/jazzy/config.rb
```diff
@@ -47,6 +48,7 @@ def initialize
       self.source_directory = Pathname.pwd
       self.excluded_files = []
       self.template_directory = Pathname(__FILE__).parent + 'templates'
+      self.assets_directory = Pathname(__FILE__).parent + '../../lib/jazzy/assets/'
```
jpsim: this should be `Pathname(__FILE__).parent + 'assets'`

gurkendoktor: thanks for your assistance, should be fixed now. sorry for all the trouble...

jpsim: no trouble at all!

lib/jazzy/doc_builder.rb
```diff
@@ -160,7 +160,7 @@ def self.write_undocumented_file(undocumented, output_dir)
     end
 
     def self.copy_assets(destination)
-      origin = Pathname(__FILE__).parent + '../../lib/jazzy/assets/.'
+      origin = options.assets_directory;
```
jpsim: we usually avoid semicolons in our ruby style, could you please remove this?

gurkendoktor: sorry, will remove it

gurkendoktor:
let me check this more thoroughly, i want the next build to be green.

jpsim:
> let me check this more thoroughly, i want the next build to be green.

You can run the tests locally after you pull down the submodules by running `bundle update` followed by `bundle exec rake spec`.

jpsim:
@gurkendoktor it'd be great if you could update this! :smile:

jpsim:
Merged in #285, thanks @gurkendoktor!

PR #265, opened 2015-07-29
[DocBuilder] Take symlinks into account when linking to github

segiddins:
Manually verified this works with Realm Swift v0.94.0

\c @jpsim


jpsim:
:+1: much nicer!

PR #269, opened 2015-07-30
Make the swift version selectable

### LINKED ISSUE #214: Support multiple Swift versions
In practice, the SourceKit XPC calls and responses don't change much (if at all), so it should be possible to support multiple Swift versions in a single jazzy version.

This will be required for good CocoaDocs integration.

segiddins:
Closes #214

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,5 +1,5 @@
 language: objective-c
-osx_image: beta-xcode6.3
+osx_image: xcode6.4
 cache: bundler
 branches:
   only:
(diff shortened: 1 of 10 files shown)


segiddins:
There's a single-line integration specs change necessary due to the CocoaPods update, but other than that this should now pass!

\c @jpsim

jpsim:
:+1: please update the integration spec. You should add a changelog entry and a sentence on this to the README too since it resolves #214.

PR #270, opened 2015-08-02
[SourceKitten] update to 0.4.4

jpsim:


lib/jazzy/SourceKitten/Rakefile
```diff
@@ -1,18 +0,0 @@
-# This Rakefile just ensures that the necessary framework symlinks are in place
-# when installed as a gem, as RubyGems doesn't play nice with symlinks.
-
-task :prepare do
```
segiddins: you sure we can get rid of this?

segiddins:
@jpsim can you summarize what the vending changes are, please?

jpsim:
> @jpsim can you summarize what the vending changes are, please?

I had mistakenly removed `lib/jazzy/SourceKitten/Rakefile` but have now added it back in with a few modifications to make it glob recursively. I've confirmed this installs properly via `bundle exec rake install` and using it to build Realm Swift docs.

The changes I made to SourceKitten's runtime search paths make it so it no longer searches for frameworks in an absolute directory like it used to (`/Library/Frameworks/`). This caused problems when installing it directly (via `make install`) vs through jazzy or Homebrew.

segiddins:
:+1:

PR #273, opened 2015-08-07
--exclude now works with relative paths

pcantrell:
The `--exclude` option currently only works with absolute paths, and fails silently with relative paths.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,19 @@
+## Master
+
+##### Breaking
+
+* None.
+
+##### Enhancements
+
+* None.
+
+##### Bug Fixes
+
+* `--exclude` now works properly if its argument is a relative path.
+  [pcantrell](https://github.com/pcantrell)
+
+
 ## 0.2.2
 
 ##### Breaking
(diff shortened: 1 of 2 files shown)


segiddins:
:+1:

pcantrell:
Obviates #247

segiddins:
@pcantrell can we just get a CHANGELOG entry? Thanks1

pcantrell:
@segiddins You bet! I’m not sure where to put it though. CHANGELOG.md on the swift-2.0 branch stops at 0.2.2, but master is already at 0.2.3. Can I add the entry if we don’t know yet what the release number for Swift 2.0 will be?

segiddins:
Ah, didn't realize you made this into the swift 2 branch. Just add a section for `Master` at the top :)

pcantrell:
OK, got it. Or a `swift-2.0` section, because it’s not on master? Or maybe not, because making it `Master` makes the merge work all pretty-like?

segiddins:
Nah, just call it `Master`, since that's what it will (eventually) be

pcantrell:
Done.

PR #275, opened 2015-08-07
Fixed broken extension? test

pcantrell:
Though it’s not clear what effect it has, the `extension?` test currently always returns false.


jpsim:
:+1: nice catch, thanks!

PR #276, opened 2015-08-07
Don’t document empty extensions

pcantrell:
Jazzy currently generates an empty doc for an extension with no visible members. Unlikes structs and classes, a memberless extension conveys no useful information.

This is a particularly bad problem because Jazzy can’t distinguish internal extensions from public ones, and thus documents _all_ extensions. The internal ones have no members, and so show up as a bunch of empty (and confusing!) pages.

Won’t work unless #275 is also merged.


segiddins:
:+1:

jpsim:
I definitely agree we shouldn't be documenting empty extensions! Thanks for this PR, and the others!

pcantrell:
Thanks to you all for writing this much-needed tool.

PR #278, opened 2015-08-08
Merge extensions into the type they extend when it is in the same project

pcantrell:
It’s common for a framework to break up a large declaration into several smaller extensions. When a framework does this with a type it declares, I consider it an implementation detail.

However, while Jazzy merges all the extensions, it keeps them separate from the core type they extend. This makes it hard to find all the methods of that type. (Take a look at how messy the Alamofire docs for `Request` come out, for example.)

This PR groups extensions with the type they extend _if_ it is declared in the same project, instead of placing them separately under “Extensions.”

I imagine this might be a controversial feature, and may also have implications I don’t understand. I’d be open to adding to this PR:

* a command-line flag to control this behavior,
* documentation, and/or
* a modification to the algorithm.

Let me know what you think.

### DIFF
--- a/lib/jazzy/assets/css/jazzy.css.scss
+++ b/lib/jazzy/assets/css/jazzy.css.scss
@@ -313,6 +313,11 @@ header {
     padding-left: 3px;
     margin-left: 15px;
   }
+  .has_default_implementation {
+    font-size: .85em;
+    color: rgba(128,128,128,1);
+    font-style: italic;
+  }
 }
 
 .pointer-container {
--- a/lib/jazzy/doc_builder.rb
+++ b/lib/jazzy/doc_builder.rb
@@ -229,6 +229,7 @@ module Jazzy
       }
       gh_token_url = gh_token_url(item, source_module)
       item_render[:github_token_url] = gh_token_url
+      item_render[:default_impl_abstract] = Jazzy.markdown.render(item.default_impl_abstract) if item.default_impl_abstract
       item_render[:return] = Jazzy.markdown.render(item.return) if item.return
       item_render[:parameters] = item.parameters if item.parameters.any?
       item_render[:url] = item.url if item.children.any?
--- a/lib/jazzy/source_declaration.rb
+++ b/lib/jazzy/source_declaration.rb
@@ -14,6 +14,7 @@ module Jazzy
     attr_accessor :name
     attr_accessor :declaration
     attr_accessor :abstract
+    attr_accessor :default_impl_abstract
     attr_accessor :discussion
     attr_accessor :return
     attr_accessor :children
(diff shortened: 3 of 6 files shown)


segiddins:
I like this change, but I wonder how @jpsim feels

pcantrell:
The main flaw as it stands is that with protocol extension, it’s not clear which methods are in the protocol (and thus required, and invoked using dynamic dispatch) and which are in the extension.

I’d prefer to do as the Apple docs do for extensions that provide a default implementation for a protocol method: instead of listing the method twice (once in the protocol and once in the extension), it just says “Default implementation” next to the protocol method.

For other protocol extensions, there should perhaps be a divider in the page or something.

All that could be a follow-up PR.

segiddins:
Ah. In that case, maybe we want to disable folding for protocol extensions until we come up with a way to make the distinction between protocol and extension methods clearer?

pcantrell:
I updated this with two protocol-centric improvements:

When an extension provides a default implementation of a protocol method, the protocol docs & the default implementation docs get merged, and the method is marked as having a default implementation.

**Example input:**

    public protocol ResourceObserver {
      /**
        Called when this observer stops observering a resource. Use for making `removeObservers(ownedBy:)` trigger
        other cleanup.
      */
      func stoppedObservingResource(resource: Resource)
    }

    public extension ResourceObserver {
      /// Does nothing.
      func stoppedObservingResource(resource: Resource) { }
    }

**Resulting docs:**

Note how both protocol and default impl docs show up.

<p align="center">default impl</p>

When an extension provides methods not required in the protocol (which won’t use dynamic dispatch, which smells like a language design mistake … but Apple says so we salute and say aye), then the methods show up in a separate “Extension members” section:

<p align="center">extension members</p>

The implementation approach I used to accomplish this is debatable. It will clobber any `MARK:` in the extension, but I don’t see a cleaner way that doesn’t involve serious remodeling.

This isn’t everything we’d want for extensions. In particular, Jazzy strips away all info about “where” conditions on extensions, and this PR doesn’t fix that. At this point, though, I think this PR is a clear improvement over current behavior.

jpsim:
This is awesome! I'm wondering if we should even do this (maybe slightly generalized) to also work for extensions on types declared externally...

jpsim:
I've cherry-picked these commits on top of master [here](https://github.com/realm/jazzy/tree/merge-extensions), which may help you bring this PR back up to date.

pcantrell:
Working on rebasing this branch now.

pcantrell:
Do you want this off of master or swift-2.0?

pcantrell:
Went off swift-2.0, because it depends on other extension-related commits on that branch.

pcantrell:
What’s involved in making the mergable? It seems like there are no objections to the functionality…?

If so, this PR will change the expected output of a bunch of the integration tests. Should someone other than me manually review those deltas (a nontrivial task!), then update the integration fixtures if the changes look good? What’s your usual process?

Issue #277, opened 2015-08-08
deduplicate_declarations merges distinct methods that should not be merged


pcantrell:
It’s unclear when and why it happens — I’m unable to reproduce it in a small test case, so it may be a SourceKit bug — but sometimes, multiple variants of a method end up with the same USR. For example, in [Siesta](https://github.com/bustoutsolutions/siesta), [these methods](https://github.com/bustoutsolutions/siesta/blob/master/Source/Resource.swift#L283-L408):

    request(_:requestMutation:)
    request(_:data:mimeType:requestMutation:)
    request(_:text:mimeType:encoding:requestMutation:)
    request(_:json:mimeType:requestMutation:)
    request(_:urlEncoded:requestMutation:)

…all end up with the USR `s:FC6Siesta8Resource7requestERR`. When this happens, `deduplicate_declarations` incorrectly merges them, and only one of them gets documented.

The original intent of `deduplicate_declarations` was to merge multiple extensions of the same entity. (See 777ba4dac97ba0dd57f1b737f2a294ccd4baae27.) It probably shouldn’t touch methods at all.

segiddins:
@pcantrell it actually is supposed to touch methods -- it's needed to play nice with documenting podspecs. The issue there is the `ERR` at the end of the USR -- that indicates generating a USR actually failed

pcantrell:
Ah, somebody else is up too late too. :)

What does the podspec case look like? As you see, I have a fix coming in a larger PR, and can account for the podspec if I know what I'm looking for.

segiddins:
The same module is documented potentially many times. I'm fairly confident that the current `deduplicate_declarations` is correct, and it's the inability to properly generate the USR that's the real issue here.

pcantrell:
In the cases where [USR, type] fails to be globally unique as Jazzy expects, [USR, type, name] works. Would that break the podspec case?

segiddins:
No it wouldn't break the podspec case, but again, this feels like a bandaide versus figuring out why you're getting error types when running SourceKitten at all.

pcantrell:
Does the `ERR` suffix come from Sourcekitten? I thought it was a SourceKit thing.

If it’s from Sourcekitten, then yes, it’s a band-aid and I’ll track down the underlying bug.

If it comes from SourceKit, then there’s an underlying Apple bug, but that doesn’t change the fact that Jazzy’s assumption of the global uniqueness of USRs is incorrect.

segiddins:
[...]

PR #279, opened 2015-08-08
Warning when USR ends with “ERR”

pcantrell:
Stops the next starry-eyed fool from filing a duplicate of #277.

### DIFF
--- a/lib/jazzy/source_declaration.rb
+++ b/lib/jazzy/source_declaration.rb
@@ -3,7 +3,8 @@ require 'jazzy/source_declaration/type'
 
 module Jazzy
   class SourceDeclaration
-    attr_accessor :type
+    attr_accessor :type      # kind of declaration (e.g. class, variable, function)
+    attr_accessor :typename  # static type of declared element (e.g. String.Type -> ())
     attr_accessor :file
     attr_accessor :line
     attr_accessor :column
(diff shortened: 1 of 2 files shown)


segiddins:
@pcantrell you're _on :fire:_ today! :+1:

pcantrell:
Little one’s sleeping soundly, spouse is off at a birthday part, and I’ve got a coding itch.

But hey, you wrote the whole dang thing!

jpsim:
This is pretty great, @pcantrell!

There are two things I'd suggest changing:

1. Instead of searching for `ERR` at the end of the USR, it'd be more appropriate to compare SourceKit's `key.typename` against `<<error type>>` (full string), since `ERR` could be anywhere in the USR and the sring `ERR` could be part of a legitimate type name.
2. The error message should suggest passing the same jazzy arguments to xcodebuild:

```
Please check for compile errors by running `xcodebuild` along with any arguments passed to jazzy's `-x` or `--xcodebuild-arguments`.
```

pcantrell:
Done. I had to add `typename` to `SourceDeclaration`, which is a little bit confusing since that class already has a `type` property that probably really should be called `kind`.

I checked the error case that brought me here, and you’re quite right: the `ERR` suffix occurs always and only when the typename is `<<error type>>`. Out of curiosity, do we know for sure that will be the case? Does Apple actually document those SourceKit keys anywhere? Or are we just snooping and hoping?

jpsim:
> Or are we just snooping and hoping?

Bingo! But it does make sense to me that if there's a single unknown component in a type, SourceKit treats the whole type as unknown.

Thanks again for the great PR!

PR #280, opened 2015-08-10
Show type declarations under heading

pcantrell:
This shows the declaration of a type under the page title, as Apple docs do (though the formatting is slightly different; see screenshot).

Useful for seeing what protocols something implements, etc.

screen shot 2015-08-10 at 4 26 15 pm


jpsim:
Yup, I dig it! But please make this target `master`, I'll then rebase `swift-2.0` to include it.

jpsim:
Merged in #284, thanks @pcantrell!

pcantrell:
Sorry I’ve been absent for all the merge fun. Thanks for getting these all in!

PR #284, opened 2015-08-26
Show type declarations under heading

jpsim:
This is #280 but targeting `master` and with updated integration specs.


segiddins:
:+1:

PR #232, opened 2015-05-15
Allow customization of copyright portion of footer

emaloney:
Pull request created [at this suggestion](https://github.com/realm/jazzy/issues/231#issuecomment-102241986)

Added four new command-line arguments:

* `--copyright_label`: Optional argument allows specifying something other than © as the default copyright label
* `--copyright_year`: Optional argument allows specifying something other than the current year as the copyright year
* `--copyright_holder`: Optional argument allows specifying something other than the author as the copyright holder
* `--copyright_url`: Optional argument allows specifying something other than the author URL as the URL linked from the copyright message

PR #286, opened 2015-08-27
support custom copyright markdown

jpsim:
successor to #232.


lib/jazzy/config.rb
```diff
@@ -62,6 +64,21 @@ def template_directory=(template_directory)
       Doc.template_path = template_directory
     end
 
+    def copyright=(copyright)
```
segiddins: I'd prefer rendering be done outside of the config class.

jpsim: moved rendering to doc.rb, it's a bit cleaner

segiddins:
👍🏻

PR #289, opened 2015-09-04
Merge extensions into the type they extend when it is in the same project

pcantrell:
Reopening #278 after accidentally triggering a Github auto-close. See that older PR for discussion.


jpsim:
So I finally took a closer look at the generated output for our integration specs diffed against the previous output and I absolutely LOVE this change! Much much better.

I also now grasp the impact of the lost `// MARK:` comments, which I'd like to look more into fixing. Also, there's a bug when extending nested types. For example, Realm has a `Configuration` struct type nested in the `Realm` class, and that has an extension for CustomStringConvertible which makes jazzy now think that the description is actually a member of `Realm` (lifted up one level).

Here's what `Classes/Realm.html` looks like:

image

and the source goes something like this:

```swift
extension Realm {
    /**
    A `Realm.Configuration` is used to describe the different options used to
    create a `Realm` instance.
    */
    public struct Configuration {
        ...
    }
}

// MARK: CustomStringConvertible

extension Realm.Configuration: CustomStringConvertible {
    /// Returns a human-readable description of the configuration.
    public var description: String {
        ...
    }
}
```

pcantrell:
Good catch on the nested types. I'll investigate when I have a chance.

Making it show both the “Default Implementation” and any `// MARK:` that applies to the extension is tricky, because I’m basically stuffing the text “Default Implementation” implementation in there as if it _were_ a `// MARK:`.

I spent half an hour looking for a better way, and didn't see one that didn't involve serious remodeling or horrible hacks. Options I pondered:

- Introduce a two-tier grouping of doc sections (which is what “Default Implementation” logically is, but this gets messy fast).
- Create a separate group that with the headline text “Default Implementation” that has some flag or dummy element that prevents the group from being pruned because it’s empty. (Hackish.)
- Explicitly model out default implementation methods as a separate attribute of the jazzy doc class. (Big impact for a small fix.)
- Stuff both “Default Implementation” and any `MARK` text into the same headline, separated by a `<br>` or something. (Also hackish.)

jpsim:
> Explicitly model out default implementation methods as a separate attribute of the jazzy doc class. (Big impact for a small fix.)

This seems like the most logical solution actually, and I don't think it would have a big impact at all.

pcantrell:
>  I don't think it would have a big impact at all.

Well, I'll get that a whirl then, and see how it plays out.

pcantrell:
Er, in my posts above, s/Default Implementation/Extension Members/.

jpsim:
@pcantrell what can I do to help move this along?

pcantrell:
@jpsim Hmm, got a time machine?

The current holdup on this one is pulling out the “Default Implementation” section as a separate thing in the doc model, instead of piggybacking on MARK. That touches lots of code, from parsing all the way to the templates (I think?).

If you want to take a look at what's involved in that model change, that would move things along. I'm certainly not going to have spare time to dig into it this week. (And I don't have any local progress on it that I'm keeping secret either.)

jpsim:
ok, time machine, on it! :wink: 

I'll see what I can do to explicitly model out default implementation methods.

pcantrell:
Looking into modeling out the extension members now.

@jpsim, how worried should I be about maintaining compatibility with peoples’ existing custom Mustache templates?

pcantrell:
I rethought how the extension members should be display, and decided it’s better to add an annotation to the side than to create a separate section. This removed the conflict with marks without any big cascade of changes.

Here’s what it looks like:

screen shot 2015-10-30 at 1 50 55 pm

That puts it in line with what this PR does for default implementations:

screen shot 2015-10-30 at 1 52 38 pm

Marks from the extension are now preserved:

screen shot 2015-10-30 at 1 51 51 pm

I’m not sure that the words “extension method” explain themselves enough. They mean two things: (1) you don’t have to implement this method to implement the protocol, and (2) you should read and understand [The Ghost of Swift Bugs Future](http://nomothetis.svbtle.com/the-ghost-of-swift-bugs-future?utm_campaign=iOS%2BDev%2BWeekly&utm_medium=web&utm_source=iOS_Dev_Weekly_Issue_203). Any thoughts on rephrasing?

paulyoung:
Glad to see this is moving forward!

Just my 2 cents: I have some projects where literally every method is an extension method, mostly for organizational purposes. Might be a use case worth considering.

Please keep up the great work!

pcantrell:
@paulyoung Have you tried running your project through this PR? Would be great to get your feedback on whether the output works well for your coding style.

paulyoung:
@pcantrell I haven't, but one is available here: https://github.com/the-grid/Disc

paulyoung:
@pcantrell to clarify – in that project, all of the methods of `APIClient` are extension methods.

pcantrell:
@jpsim The bug you mention above about nested types appears not to be related to this PR. Even in the integration specs without this PR, the `description` property that should belong to `Configuration` is [showing up on the `Realm` extension instead.](https://github.com/realm/jazzy-integration-specs/blob/c14d4041795d0ab886977729e81afd92372b4e6b/document_realm_swift/after/docs/Extensions/Realm.html#L235-L264) Clearly a bug, but not a reason to hold up this PR. Filing a separate issue for it.

jpsim:
> how worried should I be about maintaining compatibility with peoples’ existing custom Mustache templates?

I'd be surprised if many users had custom templates at the moment, but maintaining support for older templates is preferable.

jpsim:
> The bug you mention above about nested types appears not to be related to this PR.

Thanks for digging into that!

I like the "Default implementation" annotation!

pcantrell:
@paulyoung: I tried running this on your project, and it’s a huge improvement. The one big shortcoming I notice is that Jazzy doesn’t show protocols adopted by extensions — but I don’t think it did that before this PR either.

Jazzy core team, are you good with this change? What needs to be addressed to get it merged?

segiddins:
This requires an update to the integration specs? There are also a bunch of rubocop failures

pcantrell:
> This requires an update to the integration specs? There are also a bunch of rubocop failures

Right, sorry, I was unclear: I meant to say, “What _functionality_ needs to be addressed to get it merged?” I know @jpsim at least had some concerns about how it worked, and I’m wondering if it’s refined enough at this point to do the housework (specs & rubocop).

pcantrell:
Ah, I just noticed why this PR is going nuts when I rebase: it’s so old, it’s still based off of the `swift-2.0` branch.

Do I need to close & reopen this PR _again?_ Sight. Oh, git, you and your hub….

PR #290, opened 2015-09-04
Custom categories

pcantrell:
Reopening #281 after accidentally triggering a Github auto-close. See that older PR for discussion.


jpsim:
I cherry-picked these commits directly onto `swift-2.0`, so this is already merged.

Issue #244, opened 2015-05-30
"View on GitHub" header link shouldn't be generated when no GitHub URL is specified


khoogheem:
I recommend that the Link to GitHub is removed unless you specify the -g option.

OR Better yet.. I would like to actually see a option that we could specify any Project URL so if you host on GitHub or BitBucket, or want to point it back to your Company URL project page..

jpsim:
You're right that the "View on GitHub" link in the upper right of the page still gets generated regardless of if a GitHub url is specified. We'll be sure to fix that.

jpsim:
Fixed in #292.

PR #292, opened 2015-09-07
Fixed bug where "View on GitHub" was always shown

mbogh:
It seems that the bug was cause by using `to_s` on the `github_url` which caused it to be an empty string and not `null` or `false` so `mustache` would leave it out.

Refs #244 

This change naturally requires that `document_moya_podspec` is updated with a lot of `<p class="header-right"><a href="">View on GitHub</a></p>` removed.
How do we tackle this?


jpsim:
Thanks @mbogh, this is great! I just merged your accompanying integration spec PR. I'll merge this then update it to point to the appropriate commit of the integration specs repo.

mbogh:
No problem :smile:

PR #294, opened 2015-09-08
Release 0.3.0

jpsim:


pcantrell:
- We OK with exposing `--categories` in a release, when the intent is to supplant it with a config file?
- Until #289 is merged, custom categories will create duplicate entries in the nav area when a project declares an extension on its own type.

I figure the 0.x version number addresses both these concerns, but thought I should mention them.

jpsim:
I think it's fine to document the feature as-is for now, while updating it later.

PR #296, opened 2015-09-10
Added missing Swift 2 declaration types.

jpsim:


jpsim:
I also rebuilt SourceKitten using the Xcode 7 GM.

jpsim:
The dash types were chosen at a glance based on what I thought was appropriate, but may be way off.

Issue #146, opened 2015-01-08
Windows / Invalid filename


cvasilak:
Hi all,

noticed an issue when generating 'jazzy' documentation. In particular, if the html filename generated contains a ":" character such as 
```
init(baseURL:sessionConfig:requestSerializer:responseSerializer:).html
```

fails on windows (eg. checking out a project the contains this file). Is there any way this separator character can be customised in the command line to make windows happy so checking out a project won't fail?

Thank you!

jpsim:
Fixing this should simply be a matter of sanitizing the name before making a URL out of it.

mbogh:
@cvasilak, can you provide some way of replicating this outcome?

jpsim:
I believe the proper fix here is to be URL escaping the declaration name.

mbogh:
@jpsim are you thinking something as simple as adding the following to sourcekitten.rb
``` ruby
...
require 'uri'
...
              [URI.escape(doc.name) + '.html']
...

**Comment by jpsim** [comment]
it does make for fairly uglier URLs, but I believe that's the safest way to generate URLs.

**Comment by mbogh** [comment]
I'll have a go at it if I'm not overwhelmed by todays event :smile_cat:

**Comment by mbogh** [comment]
Hmm simply using `URI.escape(doc.name)` gives us `Global%20Variables.html` which makes Safari and Chrome really sad on OS X.

If we use `doc.name.gsub(/[^\w\.]/, '_')` instead we would gain filenames like `Global_Variables.html`.

Any comment on this @jpsim ?

**Comment by jpsim** [comment]
I get the feeling that this has already been solved by other ruby projects in a general way, and that `gsub(/[^\w\.]/, '_')` may not be sufficient to clean up URLs.

**Comment by mbogh** [comment]
True, so a gem like https://github.com/agilecreativity/filename_cleaner

```
require 'filename_cleaner'

[FilenameCleaner.sanitize(doc.name) + '.html']
```

Would result in `Global.Variables.html`

jpsim:
> Hmm simply using `URI.escape(doc.name)` gives us `Global%20Variables.html` which makes Safari and Chrome really sad on OS X.

I'm curious, what do you mean by this? Does it cause issues with Safari and Chrome? This seems like a simpler & better supported solution rather than `filename_cleaner`.

mbogh:
[...]

PR #304, opened 2015-09-22
Safe url naming

mbogh:
My take on #146

Requires https://github.com/realm/jazzy-integration-specs/pull/4 to pass build


lib/jazzy/source_declaration.rb
```diff
@@ -28,5 +30,9 @@ class SourceDeclaration
     def overview
       "#{abstract}\n\n#{discussion}".strip
     end
+
+    def filename
+      FilenameCleaner.sanitize("#{name}")
```
segiddins: No need for string interpolation

segiddins:
That gem seems to bring in a bunch of dependencies... Are we sure active support doesn't just have a method for this we could use?

mbogh:
As far as I have read, there is no built-in way of escaping strings so they are safe for the filesystem.
"All" filename_cleaner does is applying some regexes.

``` ruby
url = "init(baseURL:sessionConfig:requestSerializer:responseSerializer:)"
url.gsub(/[^\w\.]/, '')
 => "initbaseURLsessionConfigrequestSerializerresponseSerializer"
```

mbogh:
@segiddins do you have an opinion on the above?

segiddins:
sure, using file-cleaner is fine if @jpsim is ok with it

mbogh:
Seems that file-cleaner does not copy the input and hence changes the input.
``` ruby
head :001 > require "filename_cleaner"
 => true 
head :002 > name = "Global Variables"
 => "Global Variables" 
head :003 > new_name = FilenameCleaner.sanitize(name, '_', false)
 => "Global_Variables" 
head :004 > name
 => "Global_Variables"
```

I'm not that familiar with Ruby but it seems odd to me that I'm responsible for sending in a copy.

jpsim:
Unless I'm missing something, it does seem like `URI::escape` is a preferable solution

PR #323, opened 2015-10-24
Render special list items

### LINKED ISSUE #317: Support all Swift 2 doc keywords
Hi,
is there any possibility to treat the `throws` keyword in Swift 2 as a special case for methods; just like `returns` or the parameters.
For example:
```
/**
Description

  - throws: Error
  - returns: the list of log files
*/
public static func getLogFiles() throws -> [String] 
```
The returns description is styled in a special way just as the parameters, which are formatted with a table. It would be great to have a special style for the throws keyworks that is applied automatically for any `- throws: description`declarations.

Best regards and keep up the great work on jazzy.

jpsim:
Fixes #317. This PR should make it clear that I don't know how to SASS. I welcome suggestions on how to improve this implementation. Thoughts @orta @segiddins?

But at least it looks nice!

image

### DIFF
--- a/lib/jazzy/assets/css/jazzy.css.scss
+++ b/lib/jazzy/assets/css/jazzy.css.scss
@@ -26,6 +26,56 @@ $code_font: 0.95em Menlo, monospace;
 $gray_border: 1px solid #e2e2e2;
 $declaration_swift_border: 5px solid #cde9f4;
 
+// Asides
+$aside_attention_color: #aaa;
+$aside_attention_border: 5px solid lighten($aside_attention_color, 20%);
+$aside_author_color: #aaa;
+$aside_author_border: 5px solid lighten($aside_author_color, 20%);
+$aside_authors_color: #aaa;
+$aside_authors_border: 5px solid lighten($aside_authors_color, 20%);
+$aside_bug_color: #aaa;
+$aside_bug_border: 5px solid lighten($aside_bug_color, 20%);
+$aside_complexity_color: #aaa;
+$aside_complexity_border: 5px solid lighten($aside_complexity_color, 20%);
+$aside_copyright_color: #aaa;
+$aside_copyright_border: 5px solid lighten($aside_copyright_color, 20%);
+$aside_date_color: #aaa;
+$aside_date_border: 5px solid lighten($aside_date_color, 20%);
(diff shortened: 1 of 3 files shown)


lib/jazzy/jazzy_markdown.rb
```diff
@@ -57,15 +57,27 @@ def header(text, header_level)
...
+      %{<div class="#{type}">
+          <p class="aside-title">#{type}</p>
+          #{text}
+        </div>}
+    end
+
     def list(text, list_type)
       elided = text.gsub!(ELIDED_LI_TOKEN, '')
       return if text =~ /\A\s*\Z/ && elided
+      return text if text =~ /aside-title/
```
segiddins: this will break anything that has the literal text `aside-title` in it...

jpsim: Sure, I made it check for a bit more.

jessesquires:
Awesome! Thanks @jpsim and @segiddins! :smile: Screenshot looks great.

You are much appreciated!

img

lib/jazzy/assets/css/jazzy.css.scss
```diff
@@ -26,6 +26,56 @@ $code_font: 0.95em Menlo, monospace;
 $gray_border: 1px solid #e2e2e2;
 $declaration_swift_border: 5px solid #cde9f4;
 
+// Asides
+$aside_attention_color: #aaa;
```
orta: probably worth making `#aaa` a var too

lib/jazzy/assets/css/jazzy.css.scss
```diff
@@ -367,6 +419,248 @@ div.Swift {
   }
 }
 
+div.attention {
+  padding: 4px 0 2px 10px;
+  margin-bottom: 1em;
+  border-left: $aside_attention_border;
+  overflow-y: hidden;
+  .aside-title {
```
orta: this is basically the best bit of sass, :+1:

orta: you're using as much as I have done in my years of usage, seems :+1:

mkieselmann:
Great. Thanks @jpsim for the really really fast implementation :) Happy to use jazzy for our documentation.

PR #324, opened 2015-10-24
automatically hyperlink references to top-level declarations

jpsim:
This dead simple implementation goes a pretty long way, but there are a few cases that could be improved:

1. References in code samples are automatically hyperlinked, which doesn't always look very nice. I'm not sure what the best way to prevent this would be.
2. To make autolinking work with declarations deeper than top-level ones, we'd have to only autolink if it's on the same page (IMO).

Nevertheless, I don' think these limitations are enough to prevent adding this functionality. /cc @segiddins @orta.

### DIFF
--- a/lib/jazzy/sourcekitten.rb
+++ b/lib/jazzy/sourcekitten.rb
@@ -9,6 +9,8 @@ require 'jazzy/highlighter'
 require 'jazzy/source_declaration'
 require 'jazzy/source_mark'
 
+ELIDED_AUTOLINK_TOKEN = '36f8f5912051ae747ef441d6511ca4cb'.freeze
+
 module Jazzy
   # This module interacts with the sourcekitten command-line executable
   module SourceKitten
@@ -309,6 +311,33 @@ module Jazzy
       end.compact
     end
 
+    def self.names_and_urls(docs)
+      names_and_urls = docs.flat_map do |doc|
+        # FIXME: autolink more than just top-level items.
+        [{:name => doc.name, :url => doc.url}] # + names_and_urls(doc.children)
+      end
+    end
+
+    def self.autolink_text(text, data, url)
+      text.gsub(/\b(#{data.map { |d| Regexp.escape(d[:name]) }.join('|')})\b/) do
+        auto_url = data.find { |d| d[:name] == $1 }[:url]
+        unless auto_url == url # docs shouldn't autolink to themselves
+          "<a href=\"#{ELIDED_AUTOLINK_TOKEN}#{auto_url}\">#{$1}</a>"
+        else
+          $1
+        end
+      end
+    end
+
+    def self.autolink(docs, data)
+      docs.map do |doc|
+        doc.abstract = autolink_text(doc.abstract, data, doc.url)
+        doc.return = autolink_text(doc.return, data, doc.url) if doc.return
+        doc.children = autolink(doc.children, data)
+        doc
+      end
+    end
+
     # Parse sourcekitten STDOUT output as JSON
     # @return [Hash] structured docs
     def self.parse(sourcekitten_output, min_acl, skip_undocumented)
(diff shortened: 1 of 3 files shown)


orta:
Yeah, shippit.

lib/jazzy/sourcekitten.rb
```diff
@@ -309,6 +311,33 @@ def self.filter_excluded_files(json)
       end.compact
     end
 
+    def self.names_and_urls(docs)
+      names_and_urls = docs.flat_map do |doc|
+        # FIXME: autolink more than just top-level items.
+        [{:name => doc.name, :url => doc.url}] # + names_and_urls(doc.children)
```
segiddins: why `flat_mapping` to a single-item array?

lib/jazzy/sourcekitten.rb
```diff
@@ -309,6 +311,33 @@ def self.filter_excluded_files(json)
...
+      end
+    end
+
+    def self.autolink_text(text, data, url)
+      text.gsub(/\b(#{data.map { |d| Regexp.escape(d[:name]) }.join('|')})\b/) do
+        auto_url = data.find { |d| d[:name] == $1 }[:url]
+        unless auto_url == url # docs shouldn't autolink to themselves
```
segiddins: use `if` if there's an `else`

segiddins:
We might want to apply a more toned-down style to auto links as compared to normal links?

segiddins:
What's the performance hit on this? I can imagine it could get to be quite large...

lib/jazzy/sourcekitten.rb
jpsim: See the comment

segiddins:
> References in code samples are automatically hyperlinked, which doesn't always look very nice. I'm not sure what the best way to prevent this would be.

Maybe by doing the auto linking at an earlier stage, where we can only apply it to the sections of text we want to?

jpsim:
> What's the performance hit on this? I can imagine it could get to be quite large...

In practice, building the Xcode project takes the vast majority of the running time, so optimizing this is unlikely to have a large impact on overall run time for most projects. That being said, if you see obvious perf improvements that could be made, I'm all ears.

> Maybe by doing the auto linking at an earlier stage, where we can only apply it to the sections of text we want to?

This wouldn't help. Consider the following scenario:

```swift
public class MyClass {}

/**
 This is my function.

 ```
 This is a code sample that references MyClass.
 ```
 */
public func myFunc() {}
```

We pass the whole comment to be parsed as Markdown. We can't reliably know where the code blocks are until the Markdown is parsed, and then afterwards we could avoid autolinking references enclosed within `<code>` or `<pre>` tags, but how can you avoid greedy matches?

jpsim:
> References in code samples are automatically hyperlinked, which doesn't always look very nice. I'm not sure what the best way to prevent this would be.

The best way I can think of to do this would be to strip all URLs out of code blocks as part of `jazzy_markdown`.

lib/jazzy/sourcekitten.rb
```diff
@@ -309,6 +311,35 @@ def self.filter_excluded_files(json)
...
+        if auto_url == url # docs shouldn't autolink to themselves
+          Regexp.last_match(0)
+        else
+          "<a href=\"#{ELIDED_AUTOLINK_TOKEN}#{auto_url}\">#{name}</a>"
+        end
+      end
+    end
+
+    def self.autolink(docs, data)
+      docs.map do |doc|
```
segiddins: this should be `each`

jpsim:
> We might want to apply a more toned-down style to auto links as compared to normal links?

Good idea. What do you think of a new CSS style that only applies the blue link color on hover?

lib/jazzy/sourcekitten.rb
jpsim: I don't think so. The intent is to return an array of transformed elements, not to mutate the elements in-place. So if this is what `map` is doing (passing mutable references to the elements), then making a copy of the doc block argument would be more "pure", but functionally the same I believe.

Issue #325, opened 2015-10-24
Improve SASS


jpsim:
#323 introduced a ton of new CSS classes and SASS variables that I'm sure could be refactored to be much DRYer, but my SASS skills are lacking.

jpsim:
I think this was cleaned up enough in #326 to close this.

PR #326, opened 2015-10-28
Cleaned up aside formatting

pcantrell:
This addresses #325, at least in part.

I’m unsure how to update the integration specs. (My usual approach to git submodules is to hide until they go away.) Do I need to create a separate PR on the specs repo…? @jpsim, can you lend a hand with this if I make very sad puppy eyes?

### DIFF
--- a/lib/jazzy/assets/css/jazzy.css.scss
+++ b/lib/jazzy/assets/css/jazzy.css.scss
@@ -362,8 +362,8 @@ header {
 }
 
 .aside, .Swift {
-  padding: 0.4em 0.8em;
-  margin: 0.8em 0;
+  padding: 6px 12px;
+  margin: 12px 0;
   border-left: $aside_border;
   overflow-y: hidden;
   .aside-title {
--- a/lib/jazzy/assets/css/jazzy.css.scss
+++ b/lib/jazzy/assets/css/jazzy.css.scss
@@ -269,13 +269,6 @@ header {
   overflow-x: auto;
 }
 
-.aside-title {
-  font-size: 9px;
-  letter-spacing: 2px;
-  text-transform: uppercase;
-  padding-bottom: 0;
-}
-
 .declaration .highlight {
   overflow-x: initial; // This allows the scrollbar to show up inside declarations
   padding: 0 40px 40px 0;
@@ -367,6 +360,10 @@ header {
   border-left: $aside_border;
   overflow-y: hidden;
   .aside-title {
+    font-size: 9px;
+    letter-spacing: 2px;
+    text-transform: uppercase;
+    padding-bottom: 0;
     margin: 0;
     color: $aside_color;
     -webkit-user-select: none;
--- a/lib/jazzy/assets/css/jazzy.css.scss
+++ b/lib/jazzy/assets/css/jazzy.css.scss
@@ -371,6 +371,9 @@ header {
     color: $aside_color;
     -webkit-user-select: none;
   }
+  p:last-child {
+    margin-bottom: 0;
+  }
 }
 
 .Swift {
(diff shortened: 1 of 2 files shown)


segiddins:
👍🏻 presuming everything ends up looking the same

pcantrell:
Two visual changes:

1. Label comes out as “SEE ALSO” instead of “SEEALSO”
2. Tidied up ugly asymmetrical margins.

lib/jazzy/assets/css/jazzy.css.scss
```diff
@@ -406,258 +361,29 @@ header {
...
-    -webkit-user-select: none;
-  }
-}
-
-div.note {
-  padding: 4px 0 2px 10px;
-  margin-bottom: 1em;
-  border-left: $aside_note_border;
+.aside, .Swift {
+  padding: 0.4em 0.8em;
```
jpsim: why the change from px to em? I prefer the spacing from before.

**before**
image

**after**
image

jpsim:
Other than the margin changes, :+1: thanks!

PS: I caught this as I was rebuilding the integration fixtures, and I'm happy to do it again once this is ready to go!

lib/jazzy/assets/css/jazzy.css.scss
pcantrell: Hmm, the “before” is weirdly asymmetric for me, and my change fixed it. _Both_ of your screenshots look wrong to my eyes. Will investigate when I have a free minute….

jpsim: I agree that the "before" version isn't perfect, but better than what I generated from this PR.

pcantrell: Right. What I’m saying is that they _both_ look quite different on my machine, before as well as after, and I need to figure out why — maybe local changes….

pcantrell: Here’s what I see after the change:
screen shot 2015-10-28 at 6 26 19 pm
Note the symmetry above & below. Clearly not what you’re getting!

jpsim: That looks quite nice! Maybe I'm generating this wrong? Seems unlikely though, since I ran `rake rebuild_integration_fixtures` and the new CSS classes were being resolved.

pcantrell: Odds are good that it’s something about my hacked-up local setup. Investigating, and I’ll add a commit to the PR when I have it figured out.

pcantrell: Aha! Neither your mistake nor mine. Sometimes the aside content is inside a <p> tag, and sometimes it isn’t. You were seeing a <p> case, and I was seeing a bare text case. Just pushed a fix to make it consistent.

pcantrell:
OK, I think this is ready to go. Take a look at the new output and see how it looks to you now, @jpsim.

Do I need to have Xcode 6.x and 7.0 installed, as well as 7.1? I experimented with rebuilding the integration fixtures, but get errors about missing compilers for Swift 1.2 and 2.0.

jpsim:
Looks good now! You can point the integration specs submodule to https://github.com/realm/jazzy-integration-specs/commit/fff44b9d7bffa9add4e55eddddd26174e220f247 now.

> Do I need to have Xcode 6.x and 7.0 installed, as well as 7.1? I experimented with rebuilding the integration fixtures, but get errors about missing compilers for Swift 1.2 and 2.0.

Currently compilers for Swift 1.2 and Swift 2.0 are required to run the full test suite.

pcantrell:
Repointed integration specs and merged. First time doing that, so please LMK if I did anything wrong.

jpsim:
Hurray!

PR #337, opened 2015-11-02
[Config] Make sure the specified ACL is valid

segiddins:


segiddins:
@jpsim I'm having trouble getting Travis to load to tell me what's wrong, but this should address a bunch of issues we've gotten in the past few weeks.

segiddins:
Specs should be back to passing now!

jpsim:
Thanks Sam!

PR #341, opened 2015-11-07
Objective-C support

jpsim:
The bulk of the work can be found in https://github.com/jpsim/SourceKitten/pull/57. I still need to update the integration fixtures and add ObjC integration specs.


lib/jazzy/source_declaration/type.rb
```diff
@@ -31,6 +31,14 @@ def mark?
         kind == 'source.lang.swift.syntaxtype.comment.mark'
       end
 
+      def enum?
```
segiddins: Doesn't this clash with swift enums?

lib/jazzy/templates/task.mustache
```diff
@@ -34,7 +34,7 @@
           <div class="declaration">
             <h4>Declaration</h4>
             <div class="Swift">
```
segiddins: Class should also have the language

lib/jazzy/templates/task.mustache
```diff
@@ -1,12 +1,12 @@
...
-    <div class="task-name-container">
-      <a name="/{{uid}}"></a>
-      <a name="//apple_ref/swift/Section/{{name}}" class="dashAnchor"></a>
-      <a href="#/{{uid}}">
-        <h3 class="section-name">{{name}}</h3>
-      </a>
-    </div>
+  <div class="task-name-container">
+    <a name="/{{uid}}"></a>
+    <a name="//apple_ref/swift/Section/{{name}}" class="dashAnchor"></a>
```
segiddins: Language

lib/jazzy/sourcekitten.rb
```diff
@@ -149,6 +149,9 @@ def self.documented_child?(doc)
     def self.should_document?(doc)
       return false if doc['key.doc.comment'].to_s.include?(':nodoc:')
 
+      # Always document Objective-C declarations.
+      return true if Config.instance.objc_mode
```
segiddins: Why not check if the doc type is actually objc?

lib/jazzy/sourcekitten.rb
```diff
@@ -113,7 +113,7 @@ def self.subdir_for_doc(doc, parents)
     # Builds SourceKitten arguments based on Jazzy options
     def self.arguments_from_options(options)
       arguments = ['doc']
-      unless options.module_name.empty?
+      unless options.module_name.empty? || Config.instance.objc_mode
```
segiddins: Is this strictly necessary?

lib/jazzy/source_declaration/type.rb
```diff
@@ -72,6 +81,59 @@ def ==(other)
...
+        }.freeze,
+        'sourcekitten.source.lang.objc.decl.protocol' => {
+          jazzy: 'Protocol',
+          dash: 'Protocol',
+        }.freeze,
+        'sourcekitten.source.lang.objc.decl.typedef' => {
+          jazzy: 'Type Definition',
+          dash: 'Alias',
+        }.freeze,
+        # FIXME: this is a temporary hack to document objc
```
segiddins: Why is this a hack? What is it doing?

lib/jazzy/config.rb
```diff
@@ -84,6 +84,11 @@ class << self
                     'delete the ~/Desktop directory.'],
       default: false
 
+    config_attr :objc_mode,
+      command_line: ['--[no-]objc'],
+      description: ['Generate docs for Objective-C.'],
```
pcantrell: Braces are optional for single-item `command_line` and `description`. Not sure if it matters.

lib/jazzy/templates/task.mustache
pcantrell: This should probably be a class attr of `language` instead of `Swift` — or perhaps `class="aside language"`, which removes a special case for `.Swift` in the CSS. I tend to prefer css classes that are semantic categories, not duplicated content.

lib/jazzy/source_declaration/type.rb
jpsim: renamed to `objc_enum?`

lib/jazzy/sourcekitten.rb
jpsim: Yes, because setting the module name for an Objective-C project is still valid, it just doesn't need to be passed to SourceKitten.

lib/jazzy/source_declaration/type.rb
jpsim: It was using the same identifier as Swift Marks. Now fixed.

lib/jazzy/config.rb
jpsim: Thanks for the tip!

lib/jazzy/sourcekitten.rb
jpsim: Because at the moment, jazzy only works in a single language at a time, and lots of things will have to change to support mixed language projects, so until then I'd rather not make any assumptions as to how that future mixed support will work and keep things simpler for what we're currently doing.

jpsim:
You can browse the docs for Realm Objective-C that this branch generates here: https://static.realm.io/jazzy_demo/Realm/ (Realm PR https://github.com/realm/realm-cocoa/pull/2817).

pcantrell:
Thanks for the link to the generated docs. At a glance, the output looks impressively high-quality. I can’t comment on the implementation, but in terms of functionality, I’d say that this is at the point of “merge now and fix any bugs later.”

Also, I know it’s not this PR, but I love what the “Show on Github” links do for the docs!

jpsim:
@pcantrell thanks for taking a look! If you're curious about the implementation, most of it is in https://github.com/jpsim/SourceKitten/pull/57.

@segiddins do you know how we could ignore execution_output.txt for just the Objective-C spec using CLIntegracon? We currently have to pass in the path to the SDK, which is machine-dependent.

orta:
This looks so cool

segiddins:
Maybe just disable verbose mode for that spec?

jpsim:
> Maybe just disable verbose mode for that spec?

Does CLIntegracon *have* a verbose mode? How do I turn it off for that test?

segiddins:
Wait, doesn't jazzy have one we can disable? I can't remember....

jpsim:
> Wait, doesn't jazzy have one we can disable? I can't remember....

It's not in Jazzy. It's in CLIntegracon. If you look at the failure (https://travis-ci.org/realm/jazzy/jobs/89867901), it's CLIntegracon that's producing a different output:

```diff
-JAZZY_FAKE_DATE=YYYY-MM-DD JAZZY_FAKE_VERSION=X.X.X COCOAPODS_SKIP_UPDATE_MESSAGE=TRUE jazzy --objc --author Realm --author_url "https://realm.io" --github_url https://github.com/realm/realm-cocoa --github-file-prefix https://github.com/realm/realm-cocoa/tree/v0.96.2 --module Realm --module-version 0.96.2 --root-url https://realm.io/docs/objc/0.96.2/api/ --xcodebuild-arguments --objc,Realm/Realm.h,-x,objective-c,-isysroot,/Applications/Xcode-7.1.app/Contents/Developer/Platforms/MacOSX.platform/Developer/SDKs/MacOSX10.11.sdk,-I,/Users/jp/Projects/jazzy/spec/integration_specs/document_realm_objc/before --template-directory "docs/templates"  2>&1
+JAZZY_FAKE_DATE=YYYY-MM-DD JAZZY_FAKE_VERSION=X.X.X COCOAPODS_SKIP_UPDATE_MESSAGE=TRUE jazzy --objc --author Realm --author_url "https://realm.io" --github_url https://github.com/realm/realm-cocoa --github-file-prefix https://github.com/realm/realm-cocoa/tree/v0.96.2 --module Realm --module-version 0.96.2 --root-url https://realm.io/docs/objc/0.96.2/api/ --xcodebuild-arguments --objc,Realm/Realm.h,-x,objective-c,-isysroot,/Applications/Xcode.app/Contents/Developer/Platforms/MacOSX.platform/Developer/SDKs/MacOSX10.11.sdk,-I,/Users/travis/build/realm/jazzy/spec/integration_specs/document_realm_objc/before --template-directory "docs/templates"  2>&1
```

segiddins:
Ah, you're manually calling the command with the absolute path. Maybe just put it in back ticks and let the shell expand it?

jpsim:
> Ah, you're manually calling the command with the absolute path. Maybe just put it in back ticks and let the shell expand it?

Good idea! Done. Hopefully that works.

lib/jazzy/sourcekitten.rb
```diff
@@ -340,6 +342,18 @@ def self.autolink(docs, data)
       end
     end
 
+    def self.reject_objc_enum_typedefs(docs)
```
segiddins: What about struct typedefs?

jpsim: I've only tested this on Realm, which doesn't have any structs in its public API. So yes, this would likely have to be expanded for structs.

Repeatedly, it'd be great to add another, well-documented ObjC library. Thoughts?

segiddins:
RestKit is well documented 😉 maybe that'll be my spec for documenting an ObjC pod?

jpsim:
This looks good enough to me... merging because #YOLO

:tada: Jazzy now has Objective-C support! :tada:

pcantrell:
Crazy dances like that are what being on version 0.x is all about!

Issue #334, opened 2015-10-30
Sourcekitten does not return extension declarations


pcantrell:
For an extension of `A`, sourcekitten gives the original declaration of `A` instead of the extension declaration:

      {
        "key.kind" : "source.lang.swift.decl.extension",
        ...
        "key.annotated_decl" : "<Declaration>public final class List&lt;T : <Type usr=\"s:C10RealmSwift6Object\">Object<\/Type>&gt; : <Type usr=\"s:C10RealmSwift8ListBase\">ListBase<\/Type><\/Declaration>",
        ...

As far as I can tell, this prevents Jazzy from being able to show type constraints on extensions.

jpsim:
Fixed in #441.

Issue #333, opened 2015-10-30
Members added to extension of nested type show up in the parent


pcantrell:
Realm adds a `description` extension to the nested type `Realm.Configuration`:

https://github.com/realm/realm-cocoa/blob/master/RealmSwift-swift2.0/RealmConfiguration.swift#L168-L173

…but Jazzy puts it on `Realm:`

https://github.com/realm/jazzy-integration-specs/blob/c14d4041795d0ab886977729e81afd92372b4e6b/document_realm_swift/after/docs/Extensions/Realm.html#L235-L264

istx25:
Completed with #343.

jpsim:
This is still very much a bug: https://realm.io/docs/swift/latest/api/Classes/Realm.html#/s:vVC10RealmSwift5Realm13Configuration11descriptionSS

![image](https://cloud.githubusercontent.com/assets/474794/20637305/d1f27880-b338-11e6-8cc0-89039ac510e5.png)

istx25:
I see, sorry about that!

johnfairh:
Was hoping this would be fixed by latest sourcekitten - is not, problem seems to be that SourceKit is pretty sure that the extension is of Realm rather than Realm.Configuration.  More digging.

maximkhatskevich:
The problem persists for me as well, see this repo:
https://github.com/XCEssentials/State/tree/2.6.2

In particular, type `State` has nested type `Pending`, which is declared in extension of `State`, but member function `onUpdate` that belongs to `Pending` appears in `State` members list, not in `Pending`.

PR #343, opened 2015-11-10
Merge extension with extended type

pcantrell:
Continuing #289 based off of master instead of swift-2.0.

### DIFF
--- a/lib/jazzy/source_declaration/type.rb
+++ b/lib/jazzy/source_declaration/type.rb
@@ -65,6 +65,10 @@ module Jazzy
         kind =~ /^source\.lang\.swift\.decl\.(class|struct|protocol|enum)$/
       end
 
+      def protocol?
+        kind =~ /^source\.lang\.swift\.decl\.protocol$/
+      end
+
       def param?
         # SourceKit strangely categorizes initializer parameters as local
         # variables, so both kinds represent a parameter in jazzy.
--- a/lib/jazzy/source_declaration/type.rb
+++ b/lib/jazzy/source_declaration/type.rb
@@ -61,6 +61,10 @@ module Jazzy
         kind =~ /^source\.lang\.swift\.decl\.extension.*/
       end
 
+      def extensible?
+        kind =~ /^source\.lang\.swift\.decl\.(class|struct|protocol|enum)$/
+      end
+
       def param?
         # SourceKit strangely categorizes initializer parameters as local
         # variables, so both kinds represent a parameter in jazzy.
--- a/lib/jazzy/source_declaration/type.rb
+++ b/lib/jazzy/source_declaration/type.rb
@@ -57,16 +57,16 @@ module Jazzy
           kind.start_with?('sourcekitten.source.lang.objc.decl')
       end
 
-      def extension?
+      def swift_extension?
         kind =~ /^source\.lang\.swift\.decl\.extension.*/
       end
 
-      def extensible?
+      def swift_extensible?
         kind =~ /^source\.lang\.swift\.decl\.(class|struct|protocol|enum)$/
       end
 
-      def protocol?
-        kind =~ /^source\.lang\.swift\.decl\.protocol$/
+      def swift_protocol?
+        kind == 'source.lang.swift.decl.protocol'
       end
 
       def param?
--- a/lib/jazzy/assets/css/jazzy.css.scss
+++ b/lib/jazzy/assets/css/jazzy.css.scss
@@ -311,7 +311,7 @@ header {
     padding-left: 3px;
     margin-left: 15px;
   }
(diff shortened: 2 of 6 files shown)


pcantrell:
I went ahead and got this one ready, now that @jpsim big Obj-C push is done. It’s now based off of master instead of swift-2.0, and the specs are up to date.

This is ready to merge if you’re happy with the output. As a reminder of issues from the older discussion:

- @jpsim was concerned about clobbering `// MARK:` text with the “Default Implementation” heading. That’s now fixed.
- There was concern about #333, which turned out not to be related to this PR.
- #334 limits the usefulness of this feature somewhat, but again, that problem existed before this PR.

lib/jazzy/source_declaration/type.rb
```diff
@@ -61,6 +61,14 @@ def extension?
         kind =~ /^source\.lang\.swift\.decl\.extension.*/
       end
 
+      def extensible?
+        kind =~ /^source\.lang\.swift\.decl\.(class|struct|protocol|enum)$/
+      end
+
+      def protocol?
+        kind =~ /^source\.lang\.swift\.decl\.protocol$/
```
jpsim: it seems nicer/cleaner to just do a string comparison if we're matching for an exact string anyway?

lib/jazzy/source_declaration/type.rb
```diff
@@ -61,6 +61,14 @@ def extension?
         kind =~ /^source\.lang\.swift\.decl\.extension.*/
       end
 
+      def extensible?
```
jpsim: Now that we have both ObjC & Swift types in there, I've prefixed the other language-specific methods with `objc_` and `swift_`. Could you please do the same for `extensible?` and `protocol?`.

jpsim:
This is so so great. I'm still impressed at how concise you managed to keep this code.

:+1: other than my very minor comments.

Also, although this is something I haven't done yet, there are lots of areas now where Swift processing differs from Objective-C processing, and we should at some point go through the codepaths to branch off where needed to keep things efficient.

lib/jazzy/source_declaration/type.rb
segiddins: I don't see why we wouldn't want this new behavior to also apply in ObjC?

jpsim: I feel like Objective-C categories aren't as *lightweight* as Swift extensions. You name your Objective-C categories. Appledoc also treats them as different than their extending type.

pcantrell: Updating with `swift_` and `objc_` prefixes as you describe for this PR.

What do you think about generic tests like `protocol?` and `extensible?` plus orthogonal language tests like `objc?` and `swift?`?

pcantrell:
>This is so so great. I'm still impressed at how concise you managed to keep this code.

That’s kind of you. I actually found it a little hard to read on re-review, so I did a very light refactoring that I think helps.

@jpsim, I’ve made your very sensible fixes, so if CI likes it, we’re good to go.

lib/jazzy/source_declaration/type.rb
jpsim: > What do you think about refactoring to generic tests like protocol? and extensible? plus orthogonal language tests like objc? and swift? (after this snowballing PR is merged)?

I think that's sensible as long as they're used by both languages.

jpsim:
Looks like there are rubocop violations to address:

```
Offenses:
lib/jazzy/sourcekitten.rb:313:9: C: Align the operands of an expression in an assignment spanning multiple lines.
        .group_by { |d| deduplication_key(d) }
        ^^^^^^^^^
lib/jazzy/sourcekitten.rb:314:9: C: Align the operands of an expression in an assignment spanning multiple lines.
        .values
        ^^^^^^^
22 files inspected, 2 offenses detected
RuboCop failed!
The command "bundle exec rake spec" exited with 1.
```

pcantrell:
Yeah, weird — why didn’t those show up locally? Fixing…

pcantrell:
Not getting that rubocop failure locally. I notice that I have rubocop 0.26.1 locally, which is what’s in the Gemfile.lock, but Travis is downloading 0.34.2. ?!?

Just pushed a commit that’s a shot in the dark at what I _think_ rubocop is asking for. We’ll see if it passes.

jpsim:
Rubocop is pinned at `0.34.2` in the Gemfile.lock on master: https://github.com/realm/jazzy/blob/master/Gemfile.lock#L88

In any case, your shot in the dark passed! :tada: 

I'll add a changelog entry for you in the spirit of getting this in.

pcantrell:
GTG

pcantrell:
Oh, right , changelog. Thanks.

PR #348, opened 2015-11-11
Release 0.4.0

jpsim:
I would have liked to have CocoaPods 0.39.0 in here too, but I'm not going to add that until we can address the malformed Xcode projects issues. So it'll have to wait until the next release.


segiddins:
@jpsim I believe the CocoaPods issue has been addressed by https://github.com/CocoaPods/CocoaPods/pull/4538!

jpsim:
Nice!

PR #360, opened 2015-11-13
Fixes for broken relative path resolution

pcantrell:
This fixes two separate but similar bugs in which code incorrectly uses the working directly to resolve a path:

- Github links were generated using the path from the working dir to a specific source file, creating broken links. This fixes it so they use the path from `source_directory` to a specific source file.
- The `excluded_files` option was always resolved relative to the working directory. It is now resolved the same as other path-based options: relative to working dir when specified on the command line, and relative to config file when specified in a config file. I can also see an argument for making this always be relative to `source_directory` no matter where the option is specified. That’s tricky: on the one hand, the exception might be confusing; on the other hand, it seems like `excluded_files` will _always_ be under `source_directory`, so why would you ever want to resolve it relative to anything else?

### DIFF
--- a/lib/jazzy/config.rb
+++ b/lib/jazzy/config.rb
@@ -126,7 +126,8 @@ module Jazzy
       description: 'Files to be excluded from documentation',
       default: [],
       parse: ->(files) do
-        files.map { |f| File.expand_path(f) }
+        PathnameArray.new(
+          Array(files).map { |f| Pathname(f) })
       end
 
     config_attr :swift_version,
@@ -397,6 +398,12 @@ module Jazzy
       attr.description.each { |line| puts "  #{line}" }
       if attr.default && attr.default != ''
         puts "  Default: #{attr.default}"
+      end
+    end
+
+    class PathnameArray < Array
+      def expand_path(base_path)
+        map { |p| p.expand_path(base_path) }
       end
     end
 
--- a/lib/jazzy/config.rb
+++ b/lib/jazzy/config.rb
@@ -31,7 +31,7 @@ module Jazzy
       end
 
       def set(config, val, mark_configured: true)
-        set_raw(config, parse.call(val))
+        set_raw(config, config.instance_exec(val, &parse))
         config.method("#{name}_configured=").call(true) if mark_configured
       end
 
@@ -67,6 +67,12 @@ module Jazzy
       attr_reader :all_config_attrs
     end
 
+    attr_accessor :base_path
+
+    def expand_path(path)
+      Pathname(path).expand_path(base_path) # nil means Pathname.pwd
+    end
+
     # ──────── Build ────────
 
     # rubocop:disable Style/AlignParameters
@@ -75,7 +81,7 @@ module Jazzy
       description: 'Folder to output the HTML docs to',
       command_line: ['-o', '--output FOLDER'],
       default: 'docs',
-      parse: ->(o) { Pathname(o) }
+      parse: ->(o) { expand_path(o) }
 
     config_attr :clean,
       command_line: ['-c', '--[no-]clean'],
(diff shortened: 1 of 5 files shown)


lib/jazzy/config.rb
```diff
@@ -400,6 +401,12 @@ def print_attr_description(attr)
       end
     end
 
+    class PathnameArray < Array
```
segiddins: I really don't like introducing a new class here, when all it's doing is one `map`

pcantrell: This simplifies `expand_paths` by providing a uniform `expand_path` method for both pathnames and arrays of pathnames. This lets us just do a simple `respond_to?(:expand_path)` check:

    def expand_paths(base_path)
      self.class.all_config_attrs.each do |attr|
        val = attr.get(self)
        if val.respond_to?(:expand_path)
          attr.set_raw(self, val.expand_path(base_path))
        end
      end
    end

The alternative is a messy little case statement in that method, which I don't think would be either easier to read or less bug-prone.

segiddins: where is it that there can either be 1 pathname or an array? That sounds like a bug to me, and one that would be fixed by making sure it's always an array

pcantrell: No, it’s that we’re applying logic to _all_ the configured attributed in one fell swoop. Most of them are always single paths, but one of is always an array.

pcantrell: (The normalization you’re talking about already happens on line 130.)

segiddins: ah. I think maybe just adding a `path` attr to the config options might make sense there. That subclass is a code smell to me, it's saying we're trying to do too much magic instead of explicitly stating what should be done

pcantrell: I understand the code smell, but I don’t see the better way. What do you mean by “adding a path attr to the config options?”

segiddins: specify that the config option holds paths, getting rid of the need to manually transform everything into a pathname

pcantrell: I don’t think that’s an improvement. `PathnameArray` subclass removed, `is_path?` option added:

    def expand_paths(base_path)
      self.class.all_config_attrs.each do |attr|
        val = attr.get(self)
        if attr. is_path?
            new_val = case val
              when Array:
                val.expand_path(base_path)
              when Pathname:
                val.map { |path| path.expand_path(base_path) }
            end
            attr.set_raw(self, new_val) if new_val
          end
        end
      end
    end

That type-based case is a much stronger smell for my money.

pcantrell: (Er, had my cases backwards there, sorry. Fixed.)

jpsim:
Whatever you rubyists decide, just merge when you feel this is ready @pcantrell :+1: 

Thanks for the fixes

pcantrell:
Thanks for the go-ahead. I’m sure Sam and I will come up with something sublimely optimal between the two of us.

lib/jazzy/config.rb
pcantrell: @segiddins After sleeping on it, I came up with an alternative: https://github.com/pcantrell/jazzy/commit/cc96f9b7ba40d86b7ded80756b1bed4349a34d12

This eliminates the subclassing, suits your wish for more explicitness, and avoid the type case that I don't like.

Its big disadvantage is that there’s now a bit more statefulness on the `Config` object: whenever you set a path-based property, it’s immediately expanded based on the current value of the `base_path` instance var, which [changes and then changes back](https://github.com/pcantrell/jazzy/blob/cc96f9b7ba40d86b7ded80756b1bed4349a34d12/lib/jazzy/config.rb#L319-L330) while reading from a config file.

I don’t have strong feelings about this approach vs. the original. Let me know what you think.

segiddins: I _far_ prefer that alternative to anything I could think up!

pcantrell: OK, I’ll go with your preference, Sam! Many eyes make good code.

PR #416, opened 2015-12-12
Use Swift 2.1.1 from Xcode 7.2 by default.

nlutsenko:
Xcode 7.2 is using Swift 2.1.1 by default, making jazzy fail completely if trying to generate docs only when the latest Xcode is installed.
Tbh, didn't test the behavior of Jazzy on Swift 2.1.1, but I would assume it didn't quite change.
@jpsim can you try testing this with Realm docs?

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,7 +2,9 @@
 
 ##### Breaking
 
-* None.
+* `--swift-version` now defaults to 2.1.1 instead of 2.1.
+  [Nikita Lutsenko](https://github.com/nlutsenko)
+  [#416](https://github.com/realm/jazzy/pull/416)
 
 ##### Enhancements
 
--- a/.travis.yml
+++ b/.travis.yml
@@ -13,5 +13,5 @@ matrix:
   include:
     - osx_image: xcode6.4
       env: TRAVIS_SWIFT_VERSION=1.2
-    - osx_image: xcode7.1
-      env: TRAVIS_SWIFT_VERSION=2.1
+    - osx_image: xcode7.2
+      env: TRAVIS_SWIFT_VERSION=2.1.1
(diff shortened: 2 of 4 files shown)


jpsim:
Thanks for this @nlutsenko! You'll have to update the `.travis.yml` file:

```yaml
- osx_image: xcode7.2
  env: TRAVIS_SWIFT_VERSION=2.1.1
```

nlutsenko:
Sure, let me try that, I actually am not sure if it will succeed or not, since specs are generated for `2.1`.

jpsim:
There were very few changes from Swift 2.1 -> 2.1.1, so I'd try running the specs with 2.1.1 first to see if anything actually needs changing.

nlutsenko:
Updated stuff, let's see what happens now.

jpsim:
This looks good, but needs a changelog entry too!

nlutsenko:
Ehm, wow! It passes now! Whee!

nlutsenko:
Ok, sir, let me get right on that.

nlutsenko:
Done, let me know if you want to squash all of these.

CHANGELOG.md
```diff
@@ -2,12 +2,14 @@
...
 
-* None.
+* `--swift-version` now defaults to 2.1.1 instead of 2.1.
+  [Nikita Lutsenko](https://github.com/nlutsenko)
+  [#416](https://github.com/realm/jazzy/pull/416)
 
 ##### Enhancements
 
 * Add `--sdk [iphone|watch|appletv][os|simulator]|macosx` option for Objective-C
-  projects.
```
jpsim: All this whitespace was intentional. Markdown treats it as a line break in the rendered output.

nlutsenko: Sorry. It's TextMate, not me :P

nlutsenko: Will fix in a bit.

nlutsenko:
Should be good now.

jpsim:
Thanks again for this.

PR #436, opened 2015-12-29
Jazzercise: A flex-box based theme for jazzy

agentk:
Copied the default assets and templates folders to /themes/jazzy/
Added a simplified flex-box based version of jazzy to /themes/jazzercise/


jpsim:
Sweet! Nice to see some new styling on our old and busted theme :wink:.

Here's how I think we should proceed:

1. Move the `themes` directory to `lib/jazzy/themes`.
2. Deprecate the `--template-directory` and `--assets-directory` options, combining them into a `--theme-directory` option instead. It makes little sense to specify these separately, and it's only really like that today because they were added separately. This should be done in a separate PR.
3. Rename the `jazzy` theme to `default` or `apple`? My preference being `apple`.
4. Provide a `--theme` option which can pick from `apple` or `jazzercise`.
5. Get the theme reviewed by someone more skilled at front-end web design than I :grin:. @yoshyosh had some thoughts on this.
6. Consider renaming the `jazzercise` theme to something else. Maybe something more descriptive? `flexbox`, `full-width` or `reduxkit` come to mind.

What do you think of these steps? Would you do anything differently, or am I missing anything here?

Once we agree on the direction to take this, please let me know which parts of this you'd like to do. I'll take care of the rest :smile:

yoshyosh:
Loving the wide theme! The three biggest things that stood out to me were:
  1. Thin fonts in the headlines and side nav make it hard to read.
  2. The wide body becomes difficult to read on wider monitors. If the body was more like [this] (http://reduxkit.github.io/ReduxKit/docs/LICENSE.html) I think the readability would improve greatly!
  3. Slightly more padding above and below each header section will help readability especially as the fonts use heavier weights

agentk:
Great feedback @yoshyosh. I really like the gitbook styling, but hated having everything position fixed.

@jpsim. Good plan. I can give all those a go.
5: I'll add @yoshyosh's suggestions and leave it open for more work. 
6: I like 'fullwidth'. Giving it the cool code name of jazzercise had the desired effect of generating better suggestions.

jpsim:
:+1: :100:

agentk:
I think I've taken this about as far as I can now.

Themes are moved and renamed.

Fullwidth is pretty close to a simplified gitbook styling. Could still do with feedback again (@yoshyosh: http://reduxkit.github.io/ReduxKit/api/). I've put quite a chunk of time into cleaning up and simplifying the theme and making it readable. It will end up being the default for most of our projects in the future now.

I've set the defaults for template and assets to the apple theme.

@jpsim I might throw 2 & 4 back to you though. I wasn't sure how to deprecate the options, I couldn't see any examples of where it had been done else ware. And I wasn't sure if you intended --theme & --theme-directory to be different options that do different things. It's probably enough to just have a --theme-directory={name} option that will look in lib/jazzy/themes/{name}/ or for a path from cwd, ie, themes/snazzy/

jpsim:
I really appreciate the work that went into this @agentk! Stellar :sparkles:

I'll let @yoshyosh comment on the stylistic updates, but they look fantastic to me.

One thing I'm puzzled about: why change all the double quotes to single quotes in the mustache templates? Does that still render valid HTML? I was under the impression that HTML required double quotes for its markup.

I hear what you're saying about two options for `--theme` and `--theme-directory` being too complicated. My goal is that users can both specify a bundled theme by name (`--theme apple` or `--theme fullwidth`) *and* by directory (`--theme relative/path/to/theme`), but we can just overload `--theme` to detect if the full argument value passed matches a bundled theme name, otherwise treat the input as a path.

jpsim:
> I might throw 2 & 4 back to you though.

I'll take care of these now in a new PR.

yoshyosh:
Wow nice job, really like the readability on this, you took it much farther than I was expecting! Some minor final things that could help:

Bumping up the left nav font-size
```
.nav-group-task {
    font-size: 1em;
    list-style-type: none;
}
```

Slightly darker links
```
.nav-group-task-link {
    color: #808080;
}
```

Margin adjustments to h1, h2, h3. Currently the top and bottom margin end up making huge gaps
```
h1 {
    font-size: 2rem;
    font-weight: 700;
    margin: 1.275em 0 .6em;
}
h2 {
    font-size: 1.75rem;
    font-weight: 700;
    margin: 1.275em 0 .3em;
}

h3 {
    font-size: 1.5rem;
    font-weight: 700;
    margin: 1em 0 .3em;
}
```

Other than that :+1: from me, :clap: It looks really good!

agentk:
@yoshyosh Cheers, those tweaks are all in and live. 🎷🌟

@jpsim Oh crap 😬. I was supposed to fix those quotes. It's rolled back now, plus a couple more bugs I found. The single quotes are valid html too, but it happened because I forgot to disable the prettify on save and it was 2am. I prefer to keep quoting following existing project standards.

The single theme option as you have suggested sounds perfect. It would cover most cases I can think of.

lib/jazzy/themes/fullwidth/assets/js/jazzy.js
```diff
@@ -0,0 +1,37 @@
...
+if (typeof window.dash != 'undefined') {
+  document.documentElement.className += ' dash'
+  window.jazzy.docset = true
+}
+if (navigator.userAgent.match(/xcode/i)) {
+  document.documentElement.className += ' xcode'
+  window.jazzy.docset = true
+}
+
+// On doc load, toggle the URL hash discussion if present
```
jpsim: I think we can safely remove everything from this point on, no?

agentk: Not quite. All that serves the purpose of toggling the tokens when clicked instead of refreshing the page, and only in the browser and all tokens are displayed by default in dash.

jpsim: ok cool

jpsim:
Picked up in #444. Huge thanks for your contributions @agentk! I actually like your theme a lot more than what we currently have, we may want to consider making it the default at some point in the future, although we should let the dust settle a bit before we do that.

PR #442, opened 2015-12-30
deprecate --template-directory & --assets-directory in favor of --theme

jpsim:
This look reasonable to you @agentk?


jpsim:
From https://github.com/realm/jazzy/pull/436#issuecomment-167900903:

> It makes little sense to specify these separately, and it's only really like that today because they were added separately. This should be done in a separate PR.

agentk:
Nice! 👍I like your magic.

lib/jazzy/config.rb
```diff
@@ -247,18 +247,30 @@ def expand_path(path)
                     'Example: http://git.io/v4Bcp'],
       default: []
 
+    config_attr :theme_directory,
```
pcantrell: Note that this will make it be called “theme_directory” instead of “theme” in config files, even though it will take the same value. Might be confusing.

jpsim: yeah, but it'd also be odd to refer to it as just `theme` in the code.

PR #443, opened 2015-12-31
Add --head option to inject custom HTML into <head></head>

jpsim:
My main motivation for this is to avoid having to maintain a full fork of the theme, templates and assets in realm-cocoa, but this should be generally useful to others who want to do the same, especially injecting custom JS.


jpsim:
This will have to be updated after #444 is merged to also inject the custom head in that template.

PR #449, opened 2016-01-06
update ReduxKit link

jpsim:
The `/api` one no longer resolves. @agentk is this URL stable? If not, is there a stable URL we can point to?

### DIFF
--- a/README.md
+++ b/README.md
@@ -122,7 +122,7 @@ Two themes are provided with jazzy: `apple` (default) and `fullwidth`.
 
 Here's an example built with `apple`: <https://realm.io/docs/swift/latest/api/>
 
-Here's an example built with `fullwidth`: <http://reduxkit.github.io/ReduxKit/api/>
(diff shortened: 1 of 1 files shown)


agentk:
Yes that is the correct link now. Jazzy has taken over as the generator for all of the documentation. Not just the API anymore.

jpsim:
Hmmm, CI is failing for an unrelated version, apparently since the 0.5.0 release. I guess that messed up mustache.

jpsim:
Turns out I had forgotten to run `bundle install` after bumping the version number to 0.5.0. We should probably add a `rake update_version['x.y.z']` task to the Rakefile.

segiddins:
@jpsim just add `sh 'bundle update jazzy'` to the release task

jpsim:
The release task is defined externally (`bundler/gem_tasks` I think) and I don't know how I'd go customize that.

@segiddins If this is something you know how to do, could you please make a PR for it if you have the time? Thanks :smile:

PR #456, opened 2016-01-16
Config keys from command line opts

### LINKED ISSUE #448: 'Author' setting is ignored in config file, but works from command line?
These are the contents of my config file:

    clean: true
    swift_version: 2.1.1
    min_acl: private
    skip_undocumented: true
    readme: '../README.md'
    author: 'Nicolas Miari'
    copyright: 'Copyright © 2016 My Company. all rights reserved.'

(it is used by jazzy every time I build the docs; or so says the console. Also, jazzy failed until I updated the swift version to 2.1.1, so it _is_ being used)

My output shows the Authors list as empty (the `copyright`, though, is respected):

authorsmissing 2x

If, instead, I specify the author in the command line: 

    % jazzy --author 'Nicolas Miari'

...then it is rendered properly:

authorspresent 2x

I can't believe that something so basic has escaped everyone else until now, so I guess I'm doing something wrong...?

pcantrell:
Derives config file keys from command line options instead of property names. This results in the following changes:

- `objc_mode` → `objc`
- `config_file` → `config`
- `excluded_files` → `exclude`
- `author_name` → `author`
- `module_name` → `module`
- `version` → `module_version`
- `readme_path` → `readme`
- `custom_head` → `head`
- `theme_directory` → `theme`

This affects config files only, not command line options.

Old config file keys will no longer work, but jazzy emits a helpful warning:

    WARNING: Unknown config file attribute "module_name" (Did you mean "module"?)

That “Did you mean?” part requires a half dozen lines of extra code, which we might consider removing after this has been in place for a release or two.

This is all per a discussion thread which I now cannot find — probably because it was attached to a commit?

This may fix #448.

### DIFF
--- a/.rubocop.yml
+++ b/.rubocop.yml
@@ -74,6 +74,9 @@ GuardClause:
 Next:
   Enabled: false
 
+RegexpLiteral:
+  EnforcedStyle: mixed
+
 #- Jazzy specs -----------------------------------------------------------#
 
 # Allow for `should.match /regexp/`.
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,14 @@
+## Master
+
+##### Breaking
+
+* Config files now use the same option names as the command line. If you are
+  using one of the keys that has changed in your `.jazzy.yaml`, you will receive
+  a warning. See the [pull request](https://github.com/realm/jazzy/pull/456) for
+  a complete list of changed options. As always, you can get a list of all
+  options with `jazzy --help config`.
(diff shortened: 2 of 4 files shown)


jpsim:
This looks good to me, but should be reflected in the changelog.

jpsim:
Thanks for taking care of this, @pcantrell!

pcantrell:
Done. LMK if you want the changelog entry formatted differently.

jpsim:
Awesome! :clap:

PR #458, opened 2016-01-19
Add Siesta to specs

pcantrell:
Per a suggestion @jpsim made a while ago; finally getting around to it.

Also folded in a fix to remove all the spurious dotfiles in `after/` dirs, which weren’t being deleted by the rake task.


jpsim:
:+1: thanks Paul!

Issue #457, opened 2016-01-18
Merge Obj-C categories into interface


juliangrosshauser:
This is a request for mirroring Swift's behaviour of merging extensions with their extended type in Obj-C projects, only this time for categories and interfaces. This feature was requested for Swift in https://github.com/realm/jazzy/issues/242 and implemented in https://github.com/realm/jazzy/pull/343.

Say you have an interface `@interface MYViewController: UIViewController @end` and a category `@interface MYViewController (Category) @end`. The category documentation will end up in the `Other Categories` section, but you won't be aware of that just my looking at the interface documentation. This means you won't even notice that interface documentation is missing, unless you know that there are categories containing more documentation. This is especially an issue if your project contains lots and lots of categories and the documentation of basically one class ends up scattered in a lot of places. That's the main reason we currently can't provide jazzy documentation for PSPDFKit.

It would be a lot nicer to have the category documentation show up in the interface documentation, maybe with headers separating the categories from the interface.

steipete:
We're going to work on this feature starting Tuesday next week - shouldn't be too hard. Any help/guides/tips appreciated. (cc @esad :)

pcantrell:
Start with `merge_declarations` and `deduplication_key` in `sourcekitten.rb`. It may be as simple as scanning that family of methods for `swift_extension?`, `swift_extensible?`, etc. and replacing those with more general `extension?`, `extensible?`, etc. You’ll have to think through which behaviors should remain Swift-specific (e.g. stuff related to protocol extensions).

I intentionally limited that code to Swift when I wrote that code because I wasn’t testing it on any Obj-C code, and didn’t want stupid things to happen.

esad:
Here's my [first stab at this](https://github.com/PSPDFKit-labs/jazzy/commit/7f8d68ad0d098fe1418211f309d066009941030f) - it feels hackish though as I rely on simple replace to match the dedup key of the class being extended, because there's no reference to the extended class in the `SourceDeclaration` for the category so if we want to keep it simple it would have to work like this.

@juliangrosshauser @steipete Could you try building your docs with our version and see if it has the expected behaviour? I tested on a small objc framework and seems to behave like the swift counterpart.

esad:
@pcantrell  I've had a second take on this (see https://github.com/realm/jazzy/pull/466/) which is less hackish and tries to unify the swift/objc handling of extension deduplication. Let me know what you think!

jpsim:
Done in #466.

PR #466, opened 2016-01-29
Merge ObjC categories into docs of the extended class

esad:
This is work-in-progress on #457.

The idea was to annotate `SourceDeclaration`s representing extensions (both swift and objc) with a link to the parent declaration that is being extended, and then use this as a deduplication key. This kind of unifies the existing approach for swift extensions with new behaviour needed for objc classes/categories.

Furthermore, we found it nice to have category name appear as a mark of the merged methods.

### DIFF
--- a/lib/jazzy/source_declaration.rb
+++ b/lib/jazzy/source_declaration.rb
@@ -41,6 +41,13 @@ module Jazzy
       namespace_path.map(&:name).join('.')
     end
 
+    # If this declaration is an objc category, returns an array with the name
+    # of the extended objc class and the category name itself, i.e.
+    # ["NSString", "MyMethods"], nil otherwise.
+    def objc_category_name
+      name.split(/[\(\)]/) if type.objc_category?
+    end
+
     attr_accessor :file
     attr_accessor :line
     attr_accessor :column
--- a/lib/jazzy/sourcekitten.rb
+++ b/lib/jazzy/sourcekitten.rb
@@ -320,14 +320,18 @@ module Jazzy
     def self.deduplication_key(decl)
       if decl.type.swift_extensible? || decl.type.swift_extension?
         [decl.usr, decl.name]
+      elsif decl.type.objc_class? || decl.type.objc_category?
+        name, _ = decl.objc_category_name || decl.name
+        [name, :objc_class_and_categories]
       else
         [decl.usr, decl.name, decl.type.kind]
       end
     end
 
+    # rubocop:disable Metrics/MethodLength
     # Merges all of the given types and extensions into a single document.
     def self.merge_declarations(decls)
-      extensions, typedecls = decls.partition { |d| d.type.swift_extension? }
+      extensions, typedecls = decls.partition { |d| d.type.extension? }
 
       if typedecls.size > 1
         warn 'Found conflicting type declarations with the same name, which ' \
@@ -340,6 +344,13 @@ module Jazzy
         merge_default_implementations_into_protocol(typedecl, extensions)
         mark_members_from_protocol_extension(extensions)
         extensions.reject! { |ext| ext.children.empty? }
+      elsif typedecl && typedecl.type.objc_class?
+        # Mark children merged from categories with the name of category
+        # (unless they already have a mark)
+        extensions.each do |ext|
+          _, category_name = ext.objc_category_name
+          ext.children.each { |c| c.mark.name ||= category_name }
+        end
       end
 
       decls = typedecls + extensions
@@ -351,6 +362,7 @@ module Jazzy
         end
       end
     end
+    # rubocop:enable Metrics/MethodLength
 
     # If any of the extensions provide default implementations for methods in
     # the given protocol, merge those members into the protocol doc instead of
--- a/.rubocop.yml
+++ b/.rubocop.yml
@@ -57,6 +57,10 @@ TrailingComma:
 Style/SpecialGlobalVars:
   Enabled: false
 
+# We prefer explicit `a, _ = arr` to `a, = arr` (which could be misread as a stray comma)
+Style/TrailingUnderscoreVariable:
+  Enabled: false
+
 # For lambdas nested within certain expressions, this rule forces either ugly
 # parens or curly braces that violate the "do/end for multiline blocks" rule.
 Lambda:
--- a/lib/jazzy/source_declaration/type.rb
+++ b/lib/jazzy/source_declaration/type.rb
@@ -40,6 +40,14 @@ module Jazzy
         kind == 'sourcekitten.source.lang.objc.decl.typedef'
       end
 
+      def objc_category?
+        kind == 'sourcekitten.source.lang.objc.decl.category'
+      end
+
+      def objc_class?
+        kind == 'sourcekitten.source.lang.objc.decl.class'
+      end
+
       def swift_enum_case?
         kind == 'source.lang.swift.decl.enumcase'
       end
@@ -55,6 +63,10 @@ module Jazzy
       def declaration?
         kind.start_with?('source.lang.swift.decl') ||
           kind.start_with?('sourcekitten.source.lang.objc.decl')
+      end
+
+      def extension?
+        swift_extension? || objc_category?
       end
 
       def swift_extension?
--- a/spec/integration_spec.rb
+++ b/spec/integration_spec.rb
@@ -138,6 +138,11 @@ describe_cli 'jazzy' do
                             '--framework-root . ' \
                             "--head #{realm_head.shellescape}"
     end
+
+    describe 'Creates docs for ObjC project with a variety of contents' do
+      behaves_like cli_spec 'misc_jazzy_objc_features',
+                            '--theme fullwidth'
+    end
   end
 
   describe 'jazzy swift 2.1.1' do


lib/jazzy/sourcekitten.rb
```diff
@@ -318,8 +318,8 @@ def self.deduplicate_declarations(declarations)
 
     # Two declarations get merged if they have the same deduplication key.
     def self.deduplication_key(decl)
-      if decl.type.swift_extensible? || decl.type.swift_extension?
-        [decl.usr]
+      if (decl.type.swift_extension? || decl.type.objc_category?) && !decl.extension_parent.nil?
```
segiddins: maybe add a helper method for `extension_type?`

lib/jazzy/sourcekitten.rb
```diff
@@ -340,6 +340,14 @@ def self.merge_declarations(decls)
         merge_default_implementations_into_protocol(typedecl, extensions)
         mark_members_from_protocol_extension(extensions)
         extensions.reject! { |ext| ext.children.empty? }
+      elsif typedecl && typedecl.type.objc_class?
+        # Mark children groups coming from categories with a category name,
+        # unless they already have a mark
+        extensions.each do |ext|
+          ext.children.each do |c|
+            c.mark.name ||= (ext.extension_name || ext.name)
```
segiddins: the parens here are unnecessary

pcantrell:
Thanks for this, Esad! Two families of questions as I digest the code:

First, what’s the rationale behind computing the `extension_parent`? It looks like you’re doing it just to replace the deduplication key of `[decl.usr]` with `deduplication_key(decl.extension_parent)`; is that correct? I’m guessing that’s because Obj-C extensions come back with USRs different from the type they extend…?

If so, I wonder if perhaps the regexp dance you did in your [first crack at this](https://github.com/PSPDFKit-labs/jazzy/commit/7f8d68ad0d098fe1418211f309d066009941030f) was a better plan. It takes a _lot_ of code in this PR to construct `extension_parent`. Writing an `extract_extended_type_usr` method might be preferable.

What do those Obj-C USRs look like? Can you provide some examples?

---

Second, I wonder if it might make sense to unify Swift & Obj-C a bit more, e.g. `extension?` as a shortcut for `swift_extension? || objc_category?`,  in order to simplify some of those tangled conditionals.

I vaguely remember @jpsim saying that he has a preference for keeping Swift things and Objective-C things somewhat separate, but if he’s amenable to some judicious cross-language generalizing, it could tidy things a bit.

If we _are_ going to keep things separate, then `extension_name` should be `category_name`, because extensions are anonymous in Swift.

Nice idea using the category name as a marker, BTW.

esad:
Hi @pcantrell, thanks for your feedback!

Yes, the purpose of going through all extensions to find and set the `extension_parent` is the deduplication - it struck me as more elegant to simply say "for this extension, use the dedup key of the extended type" than to manually make it match. The `extension_name` is later used when setting the mark for the merged methods.

The USRs itself in Objective-C look like `c:objc(cy)SomeClass@SomeCategory` for the category and `c:objc(cs)SomeClass` for the class itself.

I'll try to refactor this back to a more simple solution, let's see how it looks.

esad:
I've updated the PR with a simpler approach.

segiddins:
Before merging, this will need a rebase to get rid of the merge commit

lib/jazzy/source_declaration.rb
```diff
@@ -41,6 +41,13 @@ def fully_qualified_name
       namespace_path.map(&:name).join('.')
     end
 
+    # If this declaration is an objc category, returns an array with the name
+    # of the extended objc class and the category name itself, i.e.
+    # ["NSString", "MyMethods"], nil otherwise.
+    def objc_category_name
+      type.objc_category? && name.split(/[\(\)]/) || nil
```
segiddins: This could just be `name.split(/[\(\)]/) if type.objc_category?` is

lib/jazzy/sourcekitten.rb
```diff
@@ -320,14 +320,18 @@ def self.deduplicate_declarations(declarations)
...
+      elsif decl.type.objc_class? || decl.type.objc_category?
+        name, = decl.objc_category_name || decl.name
+        [name, 'objc_class_and_categories']
       else
         [decl.usr, decl.name, decl.type.kind]
       end
     end
 
     # Merges all of the given types and extensions into a single document.
+    # rubocop:disable Metrics/MethodLength
```
segiddins: this should either be re-enabled after the method, or this should go on the same line as the method declaration

segiddins:
This warrants adding categories to an integration spec, I think

esad:
I addresses the above PR remarks, but I'm blocked by integration specs not passing on the current master. Waiting to resolve this first before I update them.

pcantrell:
I like the way this ended up. Not too complicated; things makes sense. I'm dropping a few minor notes into the code, and after those I'd say it's good to merge.

+1 to Sam’s comment that this warrants an addition to the specs, perhaps just a small addition to `misc_jazzy_features` — though I think, given this project’s somewhat loose coupling between features and tests, that could be a separate PR if it proves to be a hassle.

lib/jazzy/source_declaration.rb
```diff
@@ -41,6 +41,13 @@ def fully_qualified_name
       namespace_path.map(&:name).join('.')
     end
 
+    # If this declaration is an objc category, returns an array with the name
+    # of the extended objc class and the category name itself, i.e.
+    # ["NSString", "MyMethods"], nil otherwise.
+    def objc_category_name
+      name.split(/[\(\)]/) if type.objc_category?
```
pcantrell: The name `objc_category_name` implies that this returns a string, not an array — and that seems like the right behavior.

lib/jazzy/sourcekitten.rb
```diff
@@ -320,14 +320,18 @@ def self.deduplicate_declarations(declarations)
...
         [decl.usr, decl.name]
+      elsif decl.type.objc_class? || decl.type.objc_category?
+        name, = decl.objc_category_name || decl.name
+        [name, 'objc_class_and_categories']
       else
         [decl.usr, decl.name, decl.type.kind]
       end
     end
 
+    # rubocop:disable Metrics/MethodLength
```
pcantrell: This method might benefit from a refactoring pass in the future. Right to just disable this for now and get it out the door, though!

lib/jazzy/sourcekitten.rb
```diff
@@ -320,14 +320,18 @@ def self.deduplicate_declarations(declarations)
     def self.deduplication_key(decl)
       if decl.type.swift_extensible? || decl.type.swift_extension?
         [decl.usr, decl.name]
+      elsif decl.type.objc_class? || decl.type.objc_category?
+        name, = decl.objc_category_name || decl.name
```
pcantrell: Extraneous comma here?

lib/jazzy/sourcekitten.rb
```diff
@@ -320,14 +320,18 @@ def self.deduplicate_declarations(declarations)
...
+        [name, 'objc_class_and_categories']
```
pcantrell: I like this solution — maybe a little bit of a hack, but much tighter than manually resolving the relationships!

I'd make the second array elem a symbol.

lib/jazzy/source_declaration.rb
esad: Do you think I should split this method into two: one to return the category name and the other to return the extended class name? Or rename the above one to match the method signature?

lib/jazzy/sourcekitten.rb
esad: Yep, I was thinking of taking care of this now, but wanted to keep the scope of this PR minimal

lib/jazzy/sourcekitten.rb
segiddins: nope, the comma allows this to extract the first element of an array

esad: No, it's multiple assignment :) Rubocop complained per http://www.rubydoc.info/gems/rubocop/RuboCop/Cop/Style/TrailingUnderscoreVariable so I changed it to match the required style.

pcantrell: Stupid Rubocop. But now I see why you have objc_category_name returning two things!

lib/jazzy/source_declaration.rb
pcantrell: Now that I understand better how you’re using it, I'd suggest renaming it to make it clear that it returns two values.

esad:
@segiddins What would be the best way forward to add integration tests for the above? I tried adding some objc categories to the MiscJazzyFeatures project, but the SourceKitten ignores those (even though I added them to umbrealla header). It seems that jazzy currently doesn't support mixed Swift/ObjC projects? Should I add  a new MiscJazzyObjcFeatures project to the integration specs?

segiddins:
Yeah, I think a new `MiscJazzyObjcFeatures` project would probably be the best bet

esad:
I've added integration specs for this feature. As pointed out in the child PR, I'm bit unsure how this workflow with 2 dependent separate repos should work. Can you merge the child PR first, so that I can point to it here and we can have CI pass on this PR?

esad:
After integration specs pass I'll address the remaining code style issues.

pcantrell:
The integration specs process is a bit of a mystery to me too at times, but I merged your PR and hopefully that unblocks this merge.

As you do the last little tidying work, please feel free to edit `.rubocop.yml` to make it so you can do `name, _ =` instead of `name, =`. Anybody else feel free to pipe up in disagreement, but I find the latter terrible.

Thanks for bearing with this whole process!

esad:
@pcantrell @segiddins I've addressed all issues we discussed above, added integration specs and rebased the PR. Let me know if you feel there's something else missing.

esad:
I just wanted to let you know that we've rebuilt [PSPDFKit docs](https://pspdfkit.com/api/ios/) with this patch applied and it worked fine, with categories getting merged nicely into the class docs.

pcantrell:
I'd love to get this merged! It's only blocked on the CI failing. It looks like the integration specs are building Moya with a slightly different version of AF that the expected output recorded.

I have to speed-prep a talk I'm doing tonight, but @esad (or anyone else), if you have a chance to update the expected output, I can approve the merges.

esad:
@pcantrell As soon as https://github.com/realm/jazzy-integration-specs/pull/11/commits in jazzy-integration-specs gets merged I will update this PR

esad:
@pcantrell Looks like the specs are passing again

pcantrell:
Merging! Hooray!

Thanks for all the hard work and the patience.

jpsim:
This is great! Thanks @esad for your hard work, and @pcantrell & @segiddins for the reviews! :+1:

Issue #473, opened 2016-02-04
Order of classes not stable


esad:
I'm working on updating the integration specs for #466. However, when I run rake spec from the current jazzy master it doesn't pass because the order of classes in the generated docs seems different than the one in integration specs (see screenshots).

I'm using ruby 2.3.0p0

![screen shot 2016-02-04 at 14 53 01](https://cloud.githubusercontent.com/assets/11807/12816891/515f877a-cb4f-11e5-992b-15a54d986e1a.png)
![screen shot 2016-02-04 at 14 52 55](https://cloud.githubusercontent.com/assets/11807/12816892/51606ce4-cb4f-11e5-9c92-9f364654d167.png)

esad:
I think the cause for this is

https://github.com/realm/jazzy/blob/master/lib/jazzy/doc_builder.rb#L34

There, we first sort by `name` and then by `nav_order`. In case that `nav_order` is nil (which it is for classes in realm-objc), one would think that the first sort would remain stable, but it doesn't (at least under ruby 2.3).

Example:

```
2.3.0 :007 > %w(b z a X z Z ZZ Z).sort_by {|x| x}
 => ["X", "Z", "Z", "ZZ", "a", "b", "z", "z"] 

2.3.0 :008 > %w(b z a X z Z ZZ Z).sort_by {|x| x}.sort_by { nil }
 => ["z", "Z", "Z", "ZZ", "a", "b", "z", "X"] # <- Whops?

# Proposed fix:
2.3.0 :009 > %w(b z a X z Z ZZ Z).sort_by {|x| x}.sort_by { |x| [nil,x] }
 => ["X", "Z", "Z", "ZZ", "a", "b", "z", "z"] 
```

p2:
I came here because of a similar issue -- the first and last items in the **Classes** list are exchanged. It seems that's what's happening in your example as well (`RLMArray` should be first, `RLMSortDescriptor` last). I'll assume it's the same issue and check once this fix here has been rolled out. 👍🏼

pcantrell:
It should be fixed on master, if you want to verify the fix! (`git clone` the repo and run `bin/jazzy` from the clone.)

p2:
Yes, that does it, great!!

PR #474, opened 2016-02-04
Fix sorting (#473)

esad:
Proposed fix for sorting bug, per #473


pcantrell:
I'm surprised to learn that `sort_by` is unstable! I'm also surprised that this hasn't been causing spurious spec failures more often.

Clear bug, good fix.

jpsim:
:+1: thanks for this!

PR #453, opened 2016-01-07
support Objective-C module imports

jpsim:
Addresses #452. @rastersize could you please test this by running jazzy from source as described in our [README](https://github.com/realm/jazzy#development)?

PR #496, opened 2016-03-18
update SourceKitten to 0.11.0

jpsim:
This update picks up a buuuunch of previously missed docs: https://github.com/realm/jazzy-integration-specs/compare/skip-generation...jp-sourcekitten-0.11.0

I *think* those are all correct, and we're not mis-attributing docs, but I'd love if someone else could confirm too.

Not sure who's interested in reviewing this one /cc @pcantrell @segiddins


jpsim:
Oh, I forgot the most important feature (other than finding the missing docs)!

SourceKitten 0.11.0 dynamically links `sourcekitd.framework` at runtime based on the `xcode-select -p` setting (among others). So we should be able to remove `xcinvoke` /cc @segiddins

segiddins:
@jpsim not exactly, since xcinvoke chooses the correct Xcode from the swift version flag, which might not be the same as the selected Xcode.

jpsim:
Yeah, except I'd much rather jazzy just respect the selected Xcode version, much like any other tool that either ships with Xcode or depends on it

segiddins:
Up to you, but that's a breaking change from the current behavior.

jpsim:
Thinking about this a bit more, I'd rather have the current behavior when a Swift version is explicitly requested, and the (new) default runtime linking behavior of SourceKitten otherwise. Thanks for the feedback @segiddins!

jpsim:
Will do the update as part of #453. There should be no functional changes.

PR #510, opened 2016-04-05
Add Swift 2.2 support

tamarnachmany:
This PR provides Jazzy support to developers running Xcode 7.3. It changes all hard-coded references to Swift 2.1.1 and Xcode 7.2 to Swift 2.2 and Xcode 7.3.

Note: This is failing CI and I'm not sure why. Maybe Travis does not yet support Swift 2.2? I'm not familiar with it and use Jenkins myself. In any case, thought I would open this PR and hopefully we can sort through the CI fail and get this working for people! Also, I've never contributed here before, let me know if everything is set up correctly here.

PR #512, opened 2016-04-05
Update default Swift version to 2.2

jpsim:
Continued from #510.


jpsim:
The integration spec changes can be seen here: https://github.com/realm/jazzy-integration-specs/compare/jp-sourcekitten-0.11.0...jp-swift-2.2

PR #521, opened 2016-04-12
Remove executable flag from regular file types

mrackwitz:
That seemed odd. That might be introduced by accident?


segiddins:
👍

segiddins:
The integration spec failures are unrelated

jpsim:
:+1: thanks Marius

jpsim:
integration specs should be fixed in 9eff21d

PR #530, opened 2016-04-18
Render Swift declarations for Objective-C declarations

jpsim:
This points SourceKitten to jpsim/SourceKitten#95 in order to render the equivalent Swift declarations when generating Objective-C docs. I might not be able to revisit this for a little while, so if anyone else wants to pitch in, you're certainly welcome to do so /cc @esad @jverkoey @orta @jessesquires 

Integration specs diff is here: https://github.com/realm/jazzy-integration-specs/compare/jp-render-swift-declarations-in-objc

For some reason (probably unrelated updates to SourceKitten), some extensions docs are now missing, which needs some investigation, hence the call for help above.


lib/jazzy/doc_builder.rb
```diff
@@ -284,19 +284,20 @@ def self.render_item(item, source_module)
...
-        from_protocol_extension: item.from_protocol_extension,
-        return:                  render_markdown(item.return),
-        parameters:              (item.parameters if item.parameters.any?),
-        url:                     (item.url if item.children.any?),
-        start_line:              item.start_line,
-        end_line:                item.end_line,
+        name:                       item.name,
+        abstract:                   render_markdown(abstract),
+        declaration:                item.declaration,
+        other_language_declaration: item.other_language_declaration,
```
segiddins: to be generic, this should probably be a hash, languages to their ddeclarations

jpsim: happy to do that when we get there. for now, this is a bit simpler.

jpsim:
Here's what Objective-C docs looks like with this PR:

image

Issue #503, opened 2016-03-28
Support for excluded directories


JackoPlane:
In the same way [SwiftLint](https://github.com/realm/SwiftLint) supports whole directories to be excluded, the ability to exclude specific directories would allow for increased support for excluding things like `Vendor/`

jpsim:
@JackoPlane since Jazzy just generates documentation for modules, building your `Vendor/` code as separate frameworks would have the same effect, and would work with jazzy today. I'm fine keeping this open as a feature request, but don't see ourselves implementing this any time soon.

PR #542, opened 2016-04-25
Add ability to exclude directories in addition to files.

gurrinder:
Add ability to exclude directories in addition to files.
E.g. `exclude: ['Vendor/*']`
Related issue: https://github.com/realm/jazzy/issues/503

### DIFF
--- a/lib/jazzy/sourcekitten.rb
+++ b/lib/jazzy/sourcekitten.rb
@@ -409,7 +409,7 @@ module Jazzy
       excluded_files = Config.instance.excluded_files
       json.map do |doc|
         key = doc.keys.first
-        doc[key] unless excluded_files.detect{|file| File.fnmatch?(key, file)}
+        doc[key] unless excluded_files.detect { |f| File.fnmatch?(key, f) }
       end.compact
     end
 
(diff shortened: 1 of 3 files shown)


lib/jazzy/sourcekitten.rb
```diff
@@ -409,7 +409,7 @@ def self.filter_excluded_files(json)
       excluded_files = Config.instance.excluded_files
       json.map do |doc|
         key = doc.keys.first
-        doc[key] unless excluded_files.include?(Pathname(key))
+        doc[key] unless excluded_files.detect{|file| key.include? file.to_s}
```
segiddins: Why not use File.fnmatch ?

gurrinder: no particular reason. Just didn't think of it :) I can make the change if you'd like.

jpsim:
There are a few easy rubocop style violations that need to be addressed: https://travis-ci.org/realm/jazzy/builds/126147351

gurrinder:
Thanks @jpsim . Updated.

jpsim:
Thanks!

Issue #547, opened 2016-04-29
Jazzy doesn't escape ? in URLs


kballard:
If I have a documented implementation of the `??` operator, Jazzy creates a file `Functions/??(_:_:).html`, but it doesn't escape the `?` in hyperlinks to this function. This includes the hyperlink in the sidebar as well as the "See more" hyperlink when looking at the Functions page. The hyperlinks end up looking something like `<a href="Functions/??(_:_:).html">See more</a>`, and clicking on this hyperlink ends up showing the Functions folder in the Finder (as the `?` is interpreted as the beginning of a query string).

jazzy version: 0.6.0

jpsim:
Thanks for filing this! We really should improve our URL generation sanitization in general.

istx25:
I see that #558 needs to be reviewed and merged.

SDGGiesbrecht:
The configuration option `use_safe_filenames` and the corresponding flag `--use-safe-filenames` are now available. Until 0.7.4 is released, you can build and run from source (in your project directory):
```
git clone https://github.com/realm/jazzy
cd jazzy
bundle install
cd ..
jazzy/bin/jazzy --use-safe-filenames
rm -rf jazzy
```
Any other flags you normally use (such as `--clean`) can be added to the second last line.

Issue #361, opened 2015-11-13
Jazzy produces filenames with unsupported characters on Windows for Swift Operators


RndmTsk:
Operators such as "public func +=<K, V> (inout left: [K:V], right: [K:V]) { ... }" produce file names that include unsupported characters on Windows machines (i.e. an IIS server).

Example filename: +=(_:_:).html

RndmTsk:
Here is a list of unsupported characters: https://support.microsoft.com/en-us/kb/177506

istx25:
#558 needs to be reviewed but this issue is a little irrelevant/out of date now. Closing and will continue the conversation in that PR.

SDGGiesbrecht:
The configuration option `use_safe_filenames` and the corresponding flag `--use-safe-filenames` are now available. Until 0.7.4 is released, you can build and run from source (in your project directory):
```
git clone https://github.com/realm/jazzy
cd jazzy
bundle install
cd ..
jazzy/bin/jazzy --use-safe-filenames
rm -rf jazzy
```
Any other flags you normally use (such as `--clean`) can be added to the second last line.

PR #558, opened 2016-05-16
Escape question marks in URLs

PaulTaykalo:
Source Kitten output could have question marks in document names when parsing Swift sources. Since we cannot use question marks in URLs, we need to escape them with some adequate characters sequence. `q` was added to represent question mark, and `-` was added since no potential function name can have dashes in it

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,6 +12,9 @@
 
 ##### Bug Fixes
 
+* Escape ? in URLs.    
+  [Paul Taykalo](https://github.com/PaulTaykalo)
+  [#547](https://github.com/realm/jazzy/issues/547)
 * Uses GitHub-Flavored Markdown syntax for anchors when rendering README pages.  
   [Zachary Waldowski](https://github.com/zwaldowski)
   [#524](https://github.com/realm/jazzy/issues/524)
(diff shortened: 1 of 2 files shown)


PaulTaykalo:
This one is the fix for https://github.com/realm/jazzy/issues/547

jpsim:
Is there a way we could apply a more general URL sanitization step?

PaulTaykalo:
@jpsim for example? Are there any other issues related to the URLs sanitization?

jpsim:
Yes, there have been issues with Windows file names before too, which is why I'd like to implement a general solution to URL escaping rather than hardcoding this case, without addressing possible others.

PaulTaykalo:
@jpsim like this one https://github.com/realm/jazzy/issues/361?
It's not technically escaping. Well, its' not URL escaping, As far as I see - valid *file* name without special characters should be generated. As well as valid *URL*. Haven't seen any library/code that will do that conversion for all OSes. 
So currently there should be only these `\ / : * ? " < > |`
I would suggest to handle it in the same way, unless there's better option
What do you think?

jpsim:
That seems fine to me!

istx25:
Since the original solution proposed in this PR was rejected in favour of a more generic one, I'm going to close this PR. Feel free to re-open or file a new PR with the new/preferred solution.

Issue #64, opened 2014-11-01
Display full hierarchy in side navigation


jpsim:
Only top-level tokens are displayed in the side navigation. Nested tokens are only accessible in its parent's HTML file.

jpsim:
@yoshyosh do you think you could help out with the front-end web portion of this? My skills are definitely lacking in that area, and you're a pro!

jpsim:
This is now done, but just up to one level of nesting.

istx25:
I think it's fair to close this issue. We can re-open if necessary though.

PR #567, opened 2016-05-21
add one level of nesting in side nav

jpsim:
partly addresses #64 

Looks like this:

image

Fixes realm/realm-cocoa#3555

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,20 @@
+## Master
+
+##### Breaking
+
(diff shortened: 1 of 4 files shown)


jpsim:
integration specs diff: https://github.com/realm/jazzy-integration-specs/compare/feae9bda11666bed6dd785e3652d1373d1c3984f...7d8348b5756a1f21b0313486a361e976ade9d7f5

PR #571, opened 2016-06-01
Changelog entries for documentation and abstract options

agentk:
😊


jpsim:
We have some guidelines to keep the changelog syntax consistent: https://github.com/realm/jazzy/blob/master/CONTRIBUTING.md#tracking-changes

hard-wrapping at 80 chars and 2 spaces after the last line.

agentk:
Sweet. Seeing that, I've also changed the links to point to the issue instead of the PR then.

jpsim:
Perfect, thanks for the quick turnaround on this!

PR #582, opened 2016-06-10
Group operators by name

### LINKED ISSUE #233: Separate operators from functions
Even though Swift operators are technically functions I think it would be useful to at least optionally list operators under their own heading. Especially in [this case](http://kareman.github.io/SwiftShell/) where there is a function `$` which now looks like an operator.

segiddins:
Closes #233 
- [ ] Misc spec
- [ ] CHANGELOG

### DIFF
--- a/lib/jazzy/doc_builder.rb
+++ b/lib/jazzy/doc_builder.rb
@@ -37,7 +37,7 @@ module Jazzy
           # FIXME: include arbitrarily nested extensible types
           [{ name: child.name, url: child.url }] +
             Array(child.children.select do |sub_child|
-              sub_child.type.swift_extensible?
+              sub_child.type.swift_extensible? || child.type.overview?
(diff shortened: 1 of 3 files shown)


istx25:
Merge conflicts need to be fixed and Travis :tenis:

segiddins:
Someone needs to decide if we even want this first

jpsim:
> Someone needs to decide if we even want this first

I'm not sure myself. It would help to see this in action. Maybe on a project with lots of custom operators? Like [thoughtbot/Argo](https://github.com/thoughtbot/Argo)?

jpsim:
Thanks for the PR, Sam. I still think we need to improve how jazzy renders operators, but someone else needs to drive that exploration, and this PR hasn't progressed in the last year, so I'm closing this.

PR #585, opened 2016-06-10
fix bug where single-line declarations would include body

### LINKED ISSUE #226: Single-line computed property documentation renders source
I'm seeing some strange behavior with properties that are written on a single line. _E.g._, this:

https://github.com/stephencelis/SQLite.swift/blob/3a0435ebce11fd31e7a4176eacff5f3350e70782/SQLite/Database.swift#L33

Yields this:

screen shot 2015-05-05 at 3 25 00 pm

jpsim:
fixes #226


jpsim:
This also removes the `{ get }` or `{ get set }` in protocols, which probably isn't what we want. Thoughts?

jpsim:
Fixes #586. I've decided to merge this even with the removal of `{ get }` and `{ get set }` from interfaces and have opened #591 to track fixing that at some point. I felt the tradeoff was worth it to get this in short term.

PR #588, opened 2016-06-10
Autolink declarations within declarations and fix self-linking

jpsim:
This is all @esad's work, I just fixed a few merge conflicts with `master` and rebuilt the integration specs.

Integration specs diff: realm/jazzy-integration-specs@3159385

@esad I suspect that this will hurt performance when generating the PSPDFKit docs since your API is pretty big and that the bulk of the time spent before this was in auto-linking IIRC. It might be worth focusing on performance improvements after this is merged.


lib/jazzy/sourcekitten.rb
```diff
@@ -462,34 +478,53 @@ def self.autolink_text(text, doc, root_decls)
...
         link_target = name_traversal(parts, name_root)
 
-        if link_target && link_target.url && link_target.url != doc.url
-          "<code><a href=\"#{ELIDED_AUTOLINK_TOKEN}#{link_target.url}\">" +
-            raw_name + '</a></code>'
+        if link_target &&
+           !link_target.type.extension? &&
+           link_target.url &&
+           link_target.url != doc.url.split('#').first && # Don't link to parent
+           link_target.url != doc.url # Do not link to self
```
esad: Arrrgh, can we have this be either both "Don't" or "Do not" for consistency (my part-OCD kicking in)

jpsim: lol, this is why everyone has love/hate relationships with code linters! I'll change the other one to "Don't" too.

PR #598, opened 2016-06-23
Improved lint report readability

pcantrell:
This PR contains three improvements to make `undocumented.json` more useful & more human-readable:

- The file is now pretty printed.
- Warnings are sorted by (file, line) instead of the current somewhat haphazard ordering.
- There is a new `symbol_qualified` key that gives, for example, `"Error.Cause.UnencodableText.text"` instead of the no-so-useful `"text"`. (In a green field, I’d prefer to just replacing the existing `symbol` with that fully qualified value, but I left the old key/value intact in case anyone’s using it.)

[Here are the results.](https://github.com/realm/jazzy-integration-specs/blob/feb50ad01ce0f0f00018b4f150fd642d1cb57300/document_moya_podspec/after/docs/undocumented.json)

I know the intent was to run `undocumented.json` through some reporting tool, but it’s nice to also be able to read the file itself.

In the course of implementing this, I noticed that jazzy was generating the lint report using the raw Sourcekitten dict, which resulted in some redundant logic. I switched over to using the parsed `SourceDeclaration` object, and that cleaned up a few dozen lines of code. I put that change, the one that caused the most code disruption, in its [own commit](https://github.com/realm/jazzy/commit/9833412ea5f4eb129a3b11d2d2b1523f7402c007), and verified that it didn’t alter the integration specs. (I did, however, get [different spec output locally](https://github.com/realm/jazzy-integration-specs/commit/3b2620e6ca98d017b4ded7b1c7f3c5f9db3d154e) than was checked in even before making changes. Dependencies changed? Different Ruby? Different Sourcekitten? We’ll see what happens on CI….)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,11 @@
   [William Meleyal](https://github.com/meleyal)
   [#438](https://github.com/realm/jazzy/issues/438)
 
+* The lint report in `undocumented.json` is more human-readable: includes fully
+  qualified symbol names, pretty printed.  
+  [Paul Cantrell](https://github.com/pcantrell)
+  [#598](https://github.com/realm/jazzy/issues/598)
+
 ## 0.7.0
 
 ##### Breaking
--- a/lib/jazzy/doc_builder.rb
+++ b/lib/jazzy/doc_builder.rb
@@ -161,43 +161,15 @@ module Jazzy
       end
     end
 
-    def self.decl_for_token(token)
-      if token['key.parsed_declaration']
-        token['key.parsed_declaration']
-      elsif token['key.annotated_decl']
-        token['key.annotated_decl'].gsub(/<[^>]+>/, '')
-      elsif token['key.name']
-        token['key.name']
-      else
-        'unknown declaration'
-      end
-    end
-
-    def self.filepath_for_token(token)
-      if ENV['JAZZY_INTEGRATION_SPECS']
-        Pathname.new(token['key.filepath']).basename.to_s
-      else
-        token['key.filepath']
-      end
-    end
-
-    def self.line_number_for_token(token)
-      if token['key.doc.line']
-        token['key.doc.line'] # Objective-C
-      else
-        token['key.parsed_scope.start'] # Swift
-      end
-    end
-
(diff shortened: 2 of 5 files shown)


pcantrell:
Oh, also worth mentioning: I changed how we prevent machine-specific paths from creeping into the specs. Instead of having a “specs mode” env var change app behavior, the specs now scrub the paths from the generated output. This both cleans up app code & makes the specs more realistic. That’s what the `<TMP>` in the spec output is about.

jpsim:
Yay, @pcantrell's back! 😉 

On the whole, this looks great. 👍 whenever you get CI passing 😄

pcantrell:
Nice to be back!

OK, CI’s all sorted. Looks like it was just a Ruby 2.0 vs 2.3 thing.

Before I merge, any opinion on having the separate `symbol_qualified` key vs. just changing the value of `symbol`? i.e. this, which is what’s in the PR now:

```json
{
  "file": "/Users/paul/work/bustout/siesta/Source/Error.swift",
  "line": 200,
  "symbol": "transformer",
  "symbol_qualified": "Error.Cause.TransformerReturnedNil.transformer",
  "symbol_kind": "source.lang.swift.decl.var.instance",
  "warning": "undocumented"
}
```

vs this, which is a (small) breaking change:

```json
{
  "file": "/Users/paul/work/bustout/siesta/Source/Error.swift",
  "line": 200,
  "symbol": "Error.Cause.TransformerReturnedNil.transformer",
  "symbol_kind": "source.lang.swift.decl.var.instance",
  "warning": "undocumented"
}
```

jpsim:
I agree on changing `symbol_qualified` to `symbol`.

jpsim:
👍 feel free to merge this whenever you're ready @pcantrell!

pcantrell:
Thanks, JP. Will merge when CI passes.

(How can there be no “twiddling thumbs” emoji? …Wait! Aha!  ䷄ U+4DC4 = “hexagram for waiting”)

jpsim:
> ䷄ U+4DC4

Catchy!

pcantrell:
䷄䷄䷄䷄䷄䷄䷄䷄䷄䷄䷄

Travis during US business hours….

Issue #14, opened 2014-07-03
Search in page


jpsim:

orta:
https://pspdfkit.com/blog/2016/adding-live-search-to-jazzy/

jpsim:
Hopefully @esad can upstream those changes back in here 😄

steipete:
Yes - still scheduled. We're a bit swamped by other things but will open a PR hopefully soon.

narlei:
Search is very important to a documentation.
![screen shot 2016-07-19 at 08 14 44](https://cloud.githubusercontent.com/assets/8092922/16948531/a3c50b8e-4d8b-11e6-8bd5-9fcfc317fa3e.png)

johnfairh:
Closing, #726 to track adding to the other themes.

PR #626, opened 2016-07-29
Initial search implementation (fullwidth-only)

esad:
I wanted to get early feedback on how to best integrate the search feature (See https://github.com/realm/jazzy/issues/14) that we described in our [blog post](https://pspdfkit.com/blog/2016/adding-live-search-to-jazzy/).

This PR adds support for dumping `search.json` index if the `--searchable` configuration flag is set, as well as actually searching this index (via lunr.js) and displaying the matches in the "fullwidth" theme.

I've generated Realm ObjC docs with this branch, you can see it in action [here](https://dl.dropboxusercontent.com/u/26422/jazzysearchdemo/index.html).

Couple of questions:
- Is typeahead/lunr.js dependency fine?
- Should we turn on `--searchable` per default?
- How important is searching locally (documentation viewed via `file://` URLs)? Some browsers (Chrome) don't support XHR requests from file:// URLs so we'd have to figure out a different way of downloading the index.
- Maybe it's time to refactor the theme structure in jazzy a bit so that we don't have 2 separate copies for shared assets such as jquery, lunr etc.


lib/jazzy/search_builder.rb
```diff
@@ -0,0 +1,22 @@
+require 'tempfile'
```
segiddins: This appears to be unused

orta:
> Is typeahead/lunr.js dependency fine?

IMO, this is the only reasonable way, I'm 👍  with that

> Should we turn on --searchable per default?

I'm for it ( I'll be doing so for CocoaDocs when I can )

> How important is searching locally (documentation viewed via file:// URLs)? Some browsers (Chrome) don't support XHR requests from file:// URLs so we'd have to figure out a different way of downloading the index.

This one I'm less sure of, someone else may have a better sense, I guess the thing here is that if you're reading it via Dash or Xcode's inline reader - they have their own searches, so maybe the these can take those into account and not show the search input there?

> Maybe it's time to refactor the theme structure in jazzy a bit so that we don't have 2 separate copies for shared assets such as jquery, lunr etc.

I've read issues about React-ifiying the theme, but that might be engineering effort that might be worth saving for a separate PR

jpsim:
Merged in #725.

Issue #619, opened 2016-07-22
jazzy version: 0.7.0, can't get --exclude to ignore files


ort11:
Putting in the full file spec / path / name but jazzy is still include the file / class in the docs

Have tried /*, /filename.swift, etc.

Any help would be appreciated.

garricn:
Really enjoying our first time using jazzy, but this is also happening to us (me & @n4tr0n)

seivan:
Not just files, but also directories.
This is what we've tried with:
```yaml
exclude: ["SDK/Source/Vendored/", "SDK/Source/Vendored/*", "SDK/Source/Vendored/**/*", "SDK/Source/Vendored/*/**.swift"]
```

radutzan:
Same. Using the following statement on my .jazzy.yaml file:

    exclude:
      - Project/File.swift
      - Project/Directory/*

seivan:
Looks like https://github.com/realm/jazzy/pull/640 fix it

aasatt:
Still not able to get it to work. I am using the full path copied directly from the Project Navigator.

p2:
FYI, it didn't seem to work for me (which is why I found this), however the following works for me with jazzy 0.7.1 and a file `Sources/Keychain/Keychain.swift`:

    exclude: "**/Keychain.swift"

istx25:
Hi everyone! For some the problem has been resolved and for others it has not. This leads me to believe that there are either two (or more) problems or the original one wasn't actually fixed. Would someone (with more experience on this specific problem) be able to reproduce the problem on the latest version of Jazzy (and tools)? From there we can see what we need to do. :)

grosch:
Using version 0.7.3 I listed this in my .jazzy.yaml file:

source_directory: .
exclude: ./ExpandableDatePicker/Time\ Zone/Short/ShortTimeZoneDelegate.swift

And when I run jazzy, it still says parsing ShortTimeZoneDelegate.swift.  I've also tried without the path listed and just used the filename, but that made no difference.

grosch:
I also just tried what p2 did, using "**/*TimeZoneDelegate.swift" and that didn't work either.

ekimia:
same problem here trying to exclude a directory. Tried with the config file and the `exclude` flag

grosch:
I never got this to work, but I put the :nodoc: option in the public methods of the file I wanted excluded and that worked for me.

wuf810:
[...]

PR #640, opened 2016-08-17
Fixed wildcards in file exclude config

pcantrell:
The `exclude` config option is clearly intended to support wildcards, but they weren’t working because:

- the crucial call to `fnmatch?` passed the arguments in the wrong order and
- one of them was a Pathname instead of a string.

This PR makes wildcard excludes work as expected. The corresponding specs update adds a [regression test for wildcards](https://github.com/realm/jazzy-integration-specs/commit/570dc142cff5170572554bde320d08e4d6a709e5).

The specs were not passing when I checked them out. I just regenerated them to match what jazzy was producing without further investigation. [Here’s what changed.](https://github.com/realm/jazzy-integration-specs/commit/84836e5777c3c2588e204cb5bfaa099a94966290)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,6 +12,10 @@
   [Paul Cantrell](https://github.com/pcantrell)
   [#598](https://github.com/realm/jazzy/issues/598)
 
+* The `exclude` option now properly supports wildcards.  
+  [Paul Cantrell](https://github.com/pcantrell)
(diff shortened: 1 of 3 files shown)


pcantrell:
Once this is merged, somebody should check whether it helps with #619.

pcantrell:
YOLOOOOO

Issue #646, opened 2016-08-26
Jazzy can not parse open classes from Swift 3


rafiki270:
```
import UIKit

/**
 Main class to interface with LiveUI system
 */
open class LUI: LUIMain, LUIProtocol {


    // MARK: Accessing data

    /**
     Get image by it's key, setup in the LiveUI admin panel

     - Parameters:
     - imageKey: Image key from LiveUI admin panel
     - defaultImage: Default image that should be used if image doesn't exist. Default is nil to be returned

     - Returns: UIImage Image or nil
     */
    open static func image(_ imageKey: String, defaultImage: UIImage? = nil) -> UIImage? {
        return Images.sharedInstance.get(imageKey, defaultImage: defaultImage)
    }
``
```

picciano:
I'm confused on the status of this. I see the merge above, but doesn't look like Jazzy version has been updated and am not getting an update through gem. This is still failing with Swift 3/Xcode 8, is there something I need to do to get this to work locally?

rafiki270:
I would get we'll have to wait for the next release if you want to use gem ...

rafiki270:
also, the next big issue might be the underscore before a method variable or closure

picciano:
Guess we'll find out soon enough. Xcode 8 GM should be announced in about 15 minutes or so...

istx25:
Fixed as per #647.

Issue #645, opened 2016-08-26
Jazzy 0.7.0 not compatible with latest XCode8 betas


bgalamb:
Hi
since we upgraded from Xcode8 beta 3 to Xcode8 beta 6 we are seeing that document generation fails. Most probably because there were some changes in the inheritance modifiers e.g. open

it dies always after the last file is parsed and documentation is not generated at all.

```
Parsing XXXConstants.swift (14/14)
/Users/i040023/.jenkins/workspace/doc-resources/tools/gems/jazzy-0.7.0/lib/jazzy/source_declaration/access_control_level.rb:18:in `initialize': cannot initialize AccessControlLevel with 'source.lang.swift.accessibility.fileprivate' (RuntimeError)
        from /Users/i040023/.jenkins/workspace/doc-resources/tools/gems/jazzy-0.7.0/lib/jazzy/source_declaration/access_control_level.rb:26:in `new'
        from /Users/i040023/.jenkins/workspace/doc-resources/tools/gems/jazzy-0.7.0/lib/jazzy/source_declaration/access_control_level.rb:26:in `from_doc'
        from /Users/i040023/.jenkins/workspace/doc-resources/tools/gems/jazzy-0.7.0/lib/jazzy/sourcekitten.rb:182:in `should_document?'
        from /Users/i040023/.jenkins/workspace/doc-resources/tools/gems/jazzy-0.7.0/lib/jazzy/sourcekitten.rb:178:in `block in should_document?'
        from /Users/i040023/.jenkins/workspace/doc-resources/tools/gems/jazzy-0.7.0/lib/jazzy/sourcekitten.rb:176:in `each'
        from /Users/i040023/.jenkins/workspace/doc-resources/tools/gems/jazzy-0.7.0/lib/jazzy/sourcekitten.rb:176:in `any?'
        from /Users/i040023/.jenkins/workspace/doc-resources/tools/gems/jazzy-0.7.0/lib/jazzy/sourcekitten.rb:176:in `should_document?'
        from /Users/i040023/.jenkins/workspace/doc-resources/tools/gems/jazzy-0.7.0/lib/jazzy/sourcekitten.rb:225:in `make_doc_info'
        from /Users/i040023/.jenkins/workspace/doc-resources/tools/gems/jazzy-0.7.0/lib/jazzy/sourcekitten.rb:303:in `block in make_source_declarations'
[...]

PR #647, opened 2016-08-29
Added support for access controls of fileprivate and open

shmuelk:
Description:
  - Modified the file lib/jazzy/source_declaration/access_control_level.rb to support the new access controls of fileprivate and open, available in very current Swift 3 drivers.     

Contributors:
  - https://github.com/shmuelk
    
Issues fixed:
  - #645 
  - #646

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,7 +2,7 @@
 
 ##### Enhancements
 
-* Added support for the new access control specifiers of fileprivate and open.
+* Added support for the new access control specifiers of fileprivate and open.  
  
 ##### Bug Fixes
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,9 @@
 ##### Enhancements
 
 * Added support for the new access control specifiers of fileprivate and open.  
+  [Shmuel Kallner](https://github.com/shmuelk)
+  [#645](https://github.com/realm/jazzy/issues/645)
+  [#646](https://github.com/realm/jazzy/issues/646)
  
 ##### Bug Fixes
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,9 @@
 ## Master
 
+##### Enhancements
+
+* Added support for the new access control specifiers of fileprivate and open.
+ 
 ##### Bug Fixes
 
 * Fix issue where jazzy could not be installed from Gemfile due to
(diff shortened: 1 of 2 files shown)


pcantrell:
The code looks good to me.

The CI failure seems to be due only to Alamofire’s version changing. I’ll push a specs fix for that.

@shmuelk: You should credit yourself in the changelog!

pcantrell:
Specs update. Travis should pass if you point this PR at ba8df1. Something like this, in case you aren't yet familiar with the joy of submodules:

    cd spec/integration_specs/
    git fetch
    git checkout ba8df1
    cd -
    git add spec/integration_specs
    git commit

…and push to your fork.

shmuelk:
@pcantrell Thanks for the tip.

However the `git checkout ba8df1' fails with the message:

        error: pathspec 'ba8df1' did not match any file(s) known to git.

I looked in github and saw that the commit was ba8df10, I tried that and got the same failure.

pcantrell:
Git accepts any initial string of a hash, so `git checkout ba8df1` and `git checkout ba8df10` should both work.

Its sounds like your integration_specs submodule isn't getting the latest commit when you fetch … for whatever reason. You try this nuclear option that I use when the specs get into a funky state:

    rm -rf spec/integration_specs
    git submodule update --init --recursive

If that works, then cd back into the specs directory and try it again:

    cd spec/integration_specs/
    git checkout ba8df1
    cd -
    git add spec/integration_specs
    git commit

(Note the two `cd` calls! You are checking out a commit of the specs submodule, not the main repo.)

shmuelk:
@pcantrell Thanks for the more detailed suggestion. So far it looks better than before.

pcantrell:
Yeah, just down to Rubocop warnings. You can check those locally with `rake rubocop`.

It’s complaining that you either need to have just a single space on each side of `=`, or make all the equal signs line up. This is a spot where aligned `=` reads much better IMO, but that pushes it to an 86-char line.

@jpsim: How would you feel about switching Rubocop to a max line length of 100 instead of 80 project-wide? The 80-column limit is often a pain point for me trying to format code in Jazzy; it feels a bit tight in 2016. (My own projects I keep ~120 these days.)

pcantrell:
Looking good! Thanks for the contribution — and for bearing with all the metawork.

jpsim:
> How would you feel about switching Rubocop to a max line length of 100 instead of 80 project-wide? The 80-column limit is often a pain point for me trying to format code in Jazzy; it feels a bit tight in 2016. (My own projects I keep ~120 these days.)

👍 💯 I'm all for it

PR #653, opened 2016-09-13
Release 0.7.1

jpsim:


jpsim:
I've been falling behind on jazzy in the last month or so, but I intend to get back into it in a few weeks. Thanks for your patience and for other contributors who've stepped in 😄. Much appreciated!

jpsim:
Specs mostly pass locally, other than for Moya, which is due to upstream changes anyway, so I'm merging this now. Now waiting for Travis given how backed up it is on OS X:

image

PR #682, opened 2016-11-24
Move Code of Conduct out of README

istx25:
I prefer to have the Code of Conduct in its own file. Gives it the attention it needs rather than hiding or making people look for it. Moved it out of `README.md`, created a file `CODE_OF_CONDUCT.md` and added a link to the Code of Conduct to the top of `CONTRIBUTING.md`. Let me know what y'all think.

### DIFF
--- a/CODE_OF_CONDUCT.md
+++ b/CODE_OF_CONDUCT.md
@@ -1,1 +1,3 @@
+# Code of Conduct
+
 Jazzy conforms to the [Realm Code of Conduct](https://realm.io/conduct). By participating, contributing or using Jazzy, you are expected to uphold this code. Please report unacceptable behavior to [info@realm.io](mailto:info@realm.io).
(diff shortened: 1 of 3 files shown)


istx25:
@jpsim, pretty sure I need to rebuild the integration specs. 
Would you be able to add me to the repo?

jpsim:
I just rebuilt the integration fixtures in https://github.com/realm/jazzy-integration-specs/commit/0805a4c404af5fe13162daa6e8e768f243340c9e and updated jazzy in edef43f.

I've also given you write access to the integration specs repo.

However, I don't agree with most of the changes in this PR. Referencing the code of conduct in the readme makes it more likely that users will see it.

istx25:
Fair. I've just noticed from experience with Quick and other projects listed on the Contributor Covenant lives in its own file. I can revise the PR/changes to just include the:

>Jazzy conforms to the [Realm Code of Conduct](https://realm.io/conduct). By participating, contributing or using Jazzy, you are expected to uphold this code. Please report unacceptable behavior to [info@realm.io](mailto:info@realm.io).

Since I personally feel that this reads better than the previous statement. If it's in the README or somewhere else. Thoughts?

jpsim:
Realm has vetted the language already in use, so I'd prefer keeping it as-is. I'd welcome a PR that duplicated the conduct section of the readme into the contributing doc though. Adding another file to the root level of the repo isn't appealing to me at this point, especially since we want to refer to it from other prominent files anyway.

istx25:
Sounds good to me. I'll do that.

PR #690, opened 2016-11-29
update SourceKitten to 0.15.1

jpsim:


jpsim:
I've unfortunately uncovered a regression with this version of SourceKitten causing Objective-C enums to not be properly documented.

jpsim:
I filed an issue towards SourceKitten to track fixing that regression: jpsim/SourceKitten#304

I won't be merging this PR due to this regression, but will open a new one once the issue is resolved and another version of SourceKitten is available.

PR #698, opened 2016-12-12
Fix #429 generic type

### LINKED ISSUE #429: Generic types in functions required to be documented

jpsim:
Continued from #696.


jpsim:
@istx25 since review already happened in #696, the `github-review-needs-review` label isn't appropriate. I've changed to `github-review-accepted`.

istx25:
Sounds good to me. :)

Issue #673, opened 2016-11-10
Duplicate parameters, returns in docs


jdelaune:
Not sure if I'm formatting the docs wrong for jazzy or if there is something else going wrong. They look fine within Xcode. Here's what I have in code and this is what Jazzy produces.

![Jazzy Duplicate Params](http://chargifi-files.s3.amazonaws.com/dev/Screen%20Shot%202016-11-10%20at%2000.17.13.png)

Turns into:

![Jazzy Duplicate Params](http://chargifi-files.s3.amazonaws.com/dev/Screen%20Shot%202016-11-10%20at%2000.16.20.png)

Then also notice the floating = on the end.

![Extra =](http://chargifi-files.s3.amazonaws.com/dev/Screen%20Shot%202016-11-10%20at%2000.20.36.png)

![Extra =](http://chargifi-files.s3.amazonaws.com/dev/Screen%20Shot%202016-11-10%20at%2000.20.13.png)

Returns are also duplicating

![Extra =](http://chargifi-files.s3.amazonaws.com/dev/Screen%20Shot%202016-11-10%20at%2000.28.14.png)

![Extra =](http://chargifi-files.s3.amazonaws.com/dev/Screen%20Shot%202016-11-10%20at%2000.27.56.png)

Sorry for the jumbo images

sorenmortensen:
I'm getting the same problem as well.

My code:
<img width="555" alt="screen shot 2016-11-15 at 21 38 29" src="https://cloud.githubusercontent.com/assets/13041332/20325297/fdb0e61a-ab7b-11e6-9bde-9c30964a37a5.png">

Jazzy output:
<img width="819" alt="screen shot 2016-11-15 at 21 38 51" src="https://cloud.githubusercontent.com/assets/13041332/20325298/fdb2c390-ab7b-11e6-8b2e-7ee012f11589.png">

jdelaune:
So I've found out everything have to be lowercase and the parameters option isn't supported. You need individual parameter values.

istx25:
>So I've found out everything have to be lowercase and the parameters option isn't supported. You need individual parameter values.

Closing.

SDGGiesbrecht:
This has been annoying me for a while. @istx25, you are correct that the problem is related to casing. However, Swift’s own API Design Guidelines use capitalized forms.

See [here](https://swift.org/documentation/api-design-guidelines/#write-doc-comment) under “Write a documentation comment...”
[...]

PR #700, opened 2016-12-13
Fix 699 unsafe filenames

### LINKED ISSUE #699: Unsafe characters end up in documentation filenames.
Swift function names allow (and require in the case of the colon) characters that are dangerous to use in filenames and URLs. Jazzy naïvely uses these function names when it assigns filenames to the HTML it outputs. This results in problems when one attempts to host the documentation.

For example, the following code is valid Swift:

```swift
public func /<T>(lhs: T, rhs: T) {}
```

But the resulting file is `/(_:_:).html`, which is a disaster waiting to happen.

Test it in no time by pasting the following terminal commands. (Make sure there isn’t some other “Project” folder in the current directory first!)

```shell
rm -r Project
mkdir Project
cd Project
printf 'import PackageDescription\n\nlet package = Package(\n    name: "MyPackage"\n)' > Package.swift
mkdir Sources
cd Sources
printf '/// Some documentation\npublic func /<T>(lhs: T, rhs: T) {}' > Source.swift
cd ..
swift package generate-xcodeproj
jazzy
open docs/index.html
cd ..
```

When the documentation opens, click on the link to the function in the top left... and see the ensuing chaos.

SDGGiesbrecht:
This would fix #699, #146, #361, #547 and #558 in one fell swoop, but probably needs thorough review.

See [here for the rationale](https://github.com/realm/jazzy/issues/699#issuecomment-266633638).
See [here for a quick way to test it](https://github.com/realm/jazzy/issues/699#issue-195124872).

@jpsim, what are your thoughts?

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -19,6 +19,10 @@
   [Jeremy David Giesbrecht](https://github.com/SDGGiesbrecht)
   [#673](https://github.com/realm/jazzy/issues/673)
 
+* Fix issue where unsafe characters ended up in documentation filenames.  
+  [Jeremy David Giesbrecht](https://github.com/SDGGiesbrecht)
+  [#699](https://github.com/realm/jazzy/issues/699)
+
 ## 0.7.3
 
 ##### Breaking
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,7 +9,7 @@
 * New config option "use_safe_filenames" encodes unsafe characters when
   generating filenames. By default, documentation may receive filenames like
   "/(_:_:).html". With "use_safe_filenames", the same file will receive the name
-  "_2F_28__3A__3A_29.html" instead.  
+  "_2F_28_5F_3A_5F_3A_29.html" instead.  
   [Jeremy David Giesbrecht](https://github.com/SDGGiesbrecht)
   [#699](https://github.com/realm/jazzy/issues/699)
   [#146](https://github.com/realm/jazzy/issues/146)
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,7 +6,16 @@
 
 ##### Enhancements
 
-* None.
+* New config option "use_safe_filenames" encodes unsafe characters when
+  generating filenames. By default, documentation may receive filenames like
+  "/(_:_:).html". With "use_safe_filenames", the same file will receive the name
+  "_2F_28__3A__3A_29.html" instead.  
+  [Jeremy David Giesbrecht](https://github.com/SDGGiesbrecht)
+  [#699](https://github.com/realm/jazzy/issues/699)
+  [#146](https://github.com/realm/jazzy/issues/146)
+  [#361](https://github.com/realm/jazzy/issues/361)
+  [#547](https://github.com/realm/jazzy/issues/547)
+  [#558](https://github.com/realm/jazzy/issues/558)
 
 ##### Bug Fixes
 
@@ -18,10 +27,6 @@
 * Fix issue where parameter and return callouts were duplicated in documentation.  
   [Jeremy David Giesbrecht](https://github.com/SDGGiesbrecht)
   [#673](https://github.com/realm/jazzy/issues/673)
-
-* Fix issue where unsafe characters ended up in documentation filenames.  
-  [Jeremy David Giesbrecht](https://github.com/SDGGiesbrecht)
-  [#699](https://github.com/realm/jazzy/issues/699)
 
 ## 0.7.3
 
--- a/lib/jazzy/config.rb
+++ b/lib/jazzy/config.rb
@@ -294,6 +294,14 @@ module Jazzy
         Pathname(__FILE__).parent + 'themes' + t
       end
 
+    config_attr :use_safe_filenames,
+      command_line: '--use-safe-filenames',
+      description: 'Replace unsafe characters in filenames with an encoded '\
+                   'representation. This will reduce human readability of '\
+                   'some URLs, but may be necessary for projects that '\
+                   'expose filename-unfriendly functions such as /(_:_:)',
+      default: false
+
     config_attr :template_directory,
       command_line: ['-t', '--template-directory DIRPATH'],
       description: 'DEPRECATED: Use --theme instead.',
--- a/lib/jazzy/sourcekitten.rb
+++ b/lib/jazzy/sourcekitten.rb
@@ -70,7 +70,12 @@ module Jazzy
       docs.each do |doc|
         if !doc.parent_in_docs || doc.children.count > 0
           # Create HTML page for this doc if it has children or is root-level
-          filename_base = CGI.escape(doc.name).gsub('%', '_') # Make URL-safe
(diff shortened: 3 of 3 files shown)


SDGGiesbrecht:
“Catch URL‐illegal characters in filen**a**mes.” Someday I will learn to spell... ☺

jpsim:
This is awesome! I really appreciate the time you took to carefully consider the discussions that took place in previous issues and PRs when doing this. Looks like there are a few stylistic issues that rubocop caught. You can validate these yourself by running `bundle install` followed by `bundle exec rake rubocop`. Since this option is off by default, the integration specs don't need to be updated, so that's easy. Ping me once this is passing CI and I'll do more of a thorough review, but since this is opt-in behavior, I doubt it'll be contentious.

SDGGiesbrecht:
Thanks for the tip. I wasn’t sure how to run rubocop myself, and the path of least resistance so far was to wait for your CI suite to run it for me. ☺

SDGGiesbrecht:
It satisfies RuboCop now.

One thing that has since occurred to me is that the mapping is many‐to‐one. i.e. `/(_:_:)` and `_2F(_:_:)` will both wind up as `_2F_28__3A__3A_29`, and the two functions would get clashing filenames.

I don’t think this would ever be a real‐world issue, because any API designed with underscores and hexadecimal needs to be rethought. Also the leading underscore is normally a mark of an unwillingly public symbol, which usually ought to receive `/// :nodoc:` and never survive far enough into the program to be assigned a filename.

Nevertheless, if anyone thinks this is a real concern, the fix is simply to replace extend the line
`CGI.escape(normalized).tr('%', '_')`
with a step that manually encodes any real underscores
`CGI.escape(normalized).tr('_', '%5F').tr('%', '_')`

@jpsim do you think this should be done?

jpsim:
> do you think this should be done?

Yes, might as well make the safe URLs option generate truly safe URLs 😉

SDGGiesbrecht:
Will do.

SDGGiesbrecht:
Don’t merge yet. I want to test something first.

SDGGiesbrecht:
I should have read the documentation for `tr` before I naïvely switched it in for `gsub` like RuboCop suggested.

`CGI.escape(normalized).tr('_', '%5F').tr('%', '_')` is not at all what I wanted.
`CGI.escape(normalized).gsub('_', '%5F').tr('%', '_')` does what I intended.

By now it should be apparent that Ruby isn’t my mother tongue...

It’s fixed now though and all is well.

SDGGiesbrecht:
@jpsim, I’m considering writing some minimalist test cases for realm/jazzy-integration-specs to make sure these last few bugs don’t return (two of them didn’t affect the existing tests), but I’m not entirely sure I understand its set‐up.

If I just add to the contents of `misc_jazzy_features/before`, run `bundle exec rake rebuild_integration_fixtures` and submit a pull request, will all be well, or is additional configuration required?

jpsim:
That sounds right. To make sure we're clear (sorry if redundant):

1. Specify `use_safe_filenames` in the `misc_jazzy_features` spec's `.jazzy.yaml` file: https://github.com/realm/jazzy-integration-specs/blob/master/misc_jazzy_features/before/.jazzy.yaml
2. Run `bundle exec rake rebuild_integration_fixtures` from the root of _this_ repo, making sure that Xcode 7.3.1 is installed and `xcode-select`ed.
3. Navigate to `spec/integration_specs` to inspect the diff to make sure your change is working as expected, and that there's weren't other issues that creeped in while rebuilding the specs.
4. Make a PR towards [realm/jazzy-integration-specs](https://github.com/realm/jazzy-integration-specs) with the changes
5. Once that's merged to that repo's `master` branch, update this PR to point to the tip of `master`.

SDGGiesbrecht:
Thanks. I’ll give it a shot.

SDGGiesbrecht:
I can’t seem to get the set‐up correct. No matter how I arrange my local repositories, `bundle exec rake rebuild_integration_fixtures` just prints the following and hangs.

```
--------------------------------------------------------------------------------
Running Integration tests
--------------------------------------------------------------------------------

rm -rf spec/integration_specs/tmp
```

This has turned into a lot of work for just a unit test when the bugs have already been fixed. I don’t think I want to fight with it any further. If someone else wants to take charge (especially someone not needing to juggle forks), the following would just need to go into one of the files in [realm/jazzy-integration-specs/misc_jazzy_features/before/MiscJazzyFeatures/](https://github.com/realm/jazzy-integration-specs/tree/master/misc_jazzy_features/before/MiscJazzyFeatures) to guard against a regression.

```swift
/// Function with characters unsafe for filenames
public func /<T>(lhs: T, rhs: T) {}
```
(along with `use_safe_filenames: true` in [realm/jazzy-integration-specs/misc_jazzy_features/before/.jazzy.yaml](https://github.com/realm/jazzy-integration-specs/blob/master/misc_jazzy_features/before/.jazzy.yaml))

And the following (in the same place) would guard against a regression of #673:

```swift

/// Function (recommended documentation comment style)
///
/// - Parameters:
///     - aParameter: A parameter
///
/// - Returns: A result
public func functionA(aParameter: Bool) -> Bool {
    return true
}

/// Function (alternate documentation comment style)
///
/// - Parameter: aParameter: A parameter
public func functionB(aParameter: Bool) -> Bool {
    return true
}

/// Function (non‐standard documentation comment style)
///
/// - parameters:
///     - aParameter: A parameter
///
/// - returns: A result
public func functionC(aParameter: Bool) -> Bool {
    return true
}

/// Function (alternate non‐standard documentation comment style)
///
/// - parameter: aParameter: A parameter
public func functionD(aParameter: Bool) -> Bool {
    return true
}

```

SDGGiesbrecht:
One final note: I did notice I forgot to update the example in the changelog to match the minor change we made to the algorithm. I will do that right away.

SDGGiesbrecht:
Unless the integration tests fail, I think everything is ready from my end.

jpsim:
I'm sorry to hear you've been having trouble updating the integration specs. Are you sure you've recursively cloned the git submodules? `git submodule update --init --recursive`

I should be able to take this from here... I'll let you know if I run into any problems...

jpsim:
@SDGGiesbrecht here's the integration specs diff: https://github.com/realm/jazzy-integration-specs/commit/5abc501eb1d94100d268a49870ebc3054afce865

I've noticed that some of the links are now broken due to this change. Like clicking on the sidebar categories (e.g. "Other Classes") links to a version of the URL with spaces (`Other Classes.html`) rather than the plusses (`Other+Classes.html`).

I'm quite impressed with your level of contributions so far, so I've given you write access to both this repo and the integration specs one. Hopefully that will make it a bit easier for you to contribute, especially without having to juggle different forks and submodules pointing to different remotes 😅.

Let me know if you'd like more help with this or if you think you have it covered..

SDGGiesbrecht:
Thanks. I’ll have a look.

SDGGiesbrecht:
Continuing in #701.

SDGGiesbrecht:
Continuing in #702.

Issue #465, opened 2016-01-26
Support new doc comment fields for Swift [Xcode 7.3b2, Swift 2.2]


jessesquires:
:tada: From the release notes:

> Three new doc comment fields, namely `- keyword:`, `- recommended:` and `- recommendedover:`, allow Swift users to cooperate with code completion engine to deliver more effective code completion results. The `- keyword:` field specifies concepts that are not fully manifested in declaration names. `- recommended:` indicates other declarations are preferred to the one decorated; to the contrary, `- recommendedover:` indicates the decorated declaration is preferred to those declarations whose names are specified. (23101030)

istx25:
Any progress on this?

jpsim:
These terms just need to be added to this list: https://github.com/realm/jazzy/blob/v0.7.2/lib/jazzy/jazzy_markdown.rb#L22-L45

SDGGiesbrecht:
All current callouts are now supported. (2016‐12‐16)

jessesquires:
cc @rnystrom

PR #704, opened 2016-12-16
Updates callout list

SDGGiesbrecht:
Updated callout list to match the [list in Swift’s source](https://github.com/apple/swift/blob/master/include/swift/Markup/SimpleFields.def).

Also switched the internal list to lowercase because it is easier to map everything else to lowercase to compare it with the list than to attempt to map it to cammel case to compare it (This may one day affect SeeAlso, MutatingVariant, NonmutatingVariant, RecommendedOver and the variations of ToDo, although they aren’t handled specially yet).

This should not affect any project that does not attempt to use the new callouts.

### DIFF
--- a/lib/jazzy/jazzy_markdown.rb
+++ b/lib/jazzy/jazzy_markdown.rb
@@ -19,31 +19,39 @@ module Jazzy
       "<h#{header_level} id='#{text_slug}'>#{text}</h#{header_level}>\n"
     end
 
-    UNIQUELY_HANDLED_CALLOUTS = %w(Parameters
-                                   Parameter
-                                   Returns).freeze
-    GENERAL_CALLOUTS = %w(Attention
-                          Author
-                          Authors
-                          Bug
-                          Complexity
-                          Copyright
-                          Date
-                          Experiment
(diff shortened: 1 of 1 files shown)


SDGGiesbrecht:
Note: This fixes #465.

SDGGiesbrecht:
I would rather not put any notes in the changelog, because documentation on the semantics of these callouts is still extremely scarce.

I think it would cause less confusion if a simple note is left at #465 and that was all.

As an example of the documentation void, I can’t find anywhere thats says what form `Recommended` et al. expect their arguments to take—`function`, `function(parameter:)` or something else entirely.

jpsim:
This is amazing 😍 . Thanks!

> As an example of the documentation void, I can’t find anywhere thats says what form `Recommended` et al. expect their arguments to take—`function`, `function(parameter:)` or something else entirely.

Does Erica Sadun's book cover that? https://itunes.apple.com/us/book/swift-documentation-markup/id1049010423

---

PS: unrelated, but I want to say a huge thank you for all of your contributions and support help lately! Truly lovely to see 😊

PR #708, opened 2016-12-19
Autolink Objective-C superclass in class declaration

### LINKED ISSUE #706: Superclass in Objective-C class declaration is unlinked
The declaration for an Objective-C class has any protocols autolinked, but the superclass is never autolinked. For example, [this declaration](https://www.mapbox.com/ios-sdk/api/3.4.0-beta.5/Classes/MGLPolygon.html) reads:

```objc
@interface MGLPolygon : MGLMultiPoint <MGLOverlay>
```

“MGLMultiPoint” should link to the MGLMultiPoint class reference, but it remains unlinked. By contrast, “MGLOverlay” does link to the MGLOverlay protocol reference.

The cause appears to be that the `<span>` around “MGLMultiPoint” has a class of `nc` rather than `n` ([example of the latter](https://github.com/realm/jazzy-integration-specs/blob/a07dad280bdb94fa71efb33800d41d433967561d/document_realm_objc/after/docs/Classes/RLMResults.html#L169)).

/cc @friedbunny

1ec5:
In Objective-C class declarations, autolink the superclass if possible.

Fixes #706.

/cc @friedbunny @tmcw

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -25,6 +25,10 @@
   [Jeremy David Giesbrecht](https://github.com/SDGGiesbrecht)
   [#673](https://github.com/realm/jazzy/issues/673)
 
+* Fix issue where Objective-C superclass in declaration was unlinked.  
+  [Minh Nguyễn](https://github.com/1ec5)
+  [#706](https://github.com/realm/jazzy/issues/706)
+
 ## 0.7.3
 
 ##### Breaking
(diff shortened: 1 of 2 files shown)


jpsim:
Thanks for the PR!

Could you please add a sample exercising this bug fix in the [`misc_jazzy_objc_features`](https://github.com/realm/jazzy-integration-specs/tree/master/misc_jazzy_objc_features/before) integration spec? Fix and changelog entry both look good though! 👌

jpsim:
This looks great @1ec5, I've rebuilt the integration specs and will be merging in #713.

PR #719, opened 2017-01-03
Keep Objective-C categories of non-modular classes apart

### LINKED ISSUE #539: Different Objective-C categories on an external type get merged together
In an Objective-C project, I have two distinct categories on NSValue, but they get merged into one. The merged category takes the name of one of the original categories and now one of my `custom_categories` is incomplete. This seems to match the behavior requested in #457. Merging categories into the parent documentation makes sense as long as the parent documentation is part of the same docset. But categories on an external class, such as this Foundation class, should be kept separate.

1ec5:
Fixes #539.

/cc @pcantrell

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -40,6 +40,12 @@
   [Minh Nguyễn](https://github.com/1ec5)
   [#706](https://github.com/realm/jazzy/issues/706)
 
+* Fix issue where multiple Objective-C categories of the same external class
+  in different files were merged into one and named after the first category
+  found.  
+  [Minh Nguyễn](https://github.com/1ec5)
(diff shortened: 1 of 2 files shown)


istx25:
Hey @1ec5! Thanks for filing this PR. There are four offences that RuboCop detected. Would you be able to fix those?

1ec5:
Yes, I’ll look into them today. Ruby is regrettably one of my weaker languages, so you’re welcome to resolve the issues if they look straightforward to you.

istx25:
I'm also **very** new to Ruby but I'll try to take a look. :-) Do what you can and if we have to, I can see if someone else is available to help. 💃

jpsim:
Thanks for this @1ec5, could you please update the integration specs again now that #720 was merged?

PR #721, opened 2017-01-06
Simplify integration spec update instructions

friedbunny:
It turns out there’s a [`rake bootstrap`](https://github.com/realm/jazzy/blob/558803e7ec6bef25347a17e6550899dbaab33c52/Rakefile#L1-L18) command that manages `bundle install` and submodule initialization, so I’ve gone ahead and updated the spec update instructions to use that.

/cc @captainbarbosa


jpsim:
Thanks!

Issue #714, opened 2016-12-24
Quotation marks in code blocks are converted to <q> tags


1ec5:
jazzy automatically converts quotation marks inside code blocks into `<q>` tags. Though the code looks more sophisticated as a result, it doesn’t necessarily compile. For example, see the Mapbox iOS SDK’s [MGLCircleStyleLayer](https://www.mapbox.com/ios-sdk/api/3.4.0-beta.6/Classes/MGLCircleStyleLayer.html) documentation, which places a three-backtick code block in the class’s description to provide example code:

![MGLCircleStyleLayer](https://cloud.githubusercontent.com/assets/1231218/21464364/d92b9296-c92f-11e6-96de-a1e25c914c73.png)

If you copy-paste this example from Firefox, Firefox inserts straight quotation marks around each `<q>` tag, so the example works just fine. However, in Safari and Chrome, no quotation marks are inserted, so the code doesn’t compile.

This is part of the problem in #630, but there seems to be more going on there than just the `<q>` tags.

/cc @friedbunny @ericrwolfe

1ec5:
This looks to be some kind of weird interaction between rouge and redcarpet, but I can’t quite figure out what the issue is.

In the meantime, I came up with a workaround in mapbox/mapbox-gl-native#7629 that turns the `<q>` tags into plain text on page load using JavaScript. That PR is specific to the theme we’re using for the Mapbox iOS/macOS SDK documentation. Would you accept a PR that works around the issue in similar fashion in the apple and fullwidth themes?

jpsim:
Yeah, I think that's a worthwhile hack to get in place for others to use.

johnfairh:
Underlying problem fixed by #628, workaround now removed in master.

PR #737, opened 2017-01-22
Replace <q> tags with quotation marks in code blocks

1ec5:
Work around #714 by replacing `<q>` tags inside `<code>` tags with ASCII double quotation marks in JavaScript.

Upstreamed from mapbox/mapbox-gl-native#7629.

/cc @friedbunny


1ec5:
Changelog blurb added in 97d225029c29fd24f085f30b5e58ce94ed3c5aac.

jpsim:
Thanks for this 😄

Issue #630, opened 2016-08-01
Example Swift code can not contain string interpolation


blochberger:
Documenting a function providing example code renders incorrectly if the example code contains string interpolation:

`````` swift
/**
    #### Example
    ```swift
    let result = doSomething()
    print("Result: \(result)")
    ```
*/
func doSomething() -> String {
    …
}
``````

Which will render as:

> **Example**
> 
> ``` swift
> let result = doSomething()
> print(“Result: ”se“>\(”n“>result”se“>)”s“>”)
> ```

Where the quotation is done by HTML quotes:

``` html
<span class="nf">print</span><span class="p">(</span><span class="s"><q>Result: </q></span><span class="&lt;/q">se<q>&gt;\(</q></span><span class="&lt;/q">n<q>&gt;result</q></span><span class="&lt;/q">se<q>&gt;)</q></span><span class="&lt;/q">s<q>&gt;</q></span><span class="p">)</span>
</code>
```

This issue does not arise when using i.e. `README.md` as abstract. There the quotation marks are rendered as characters as appearing in the source code `"`.

This might be related to #628.

istx25:
I believe this is related to #628. Closing as a duplicate. Will reference this issue in the future where necessary.

jpsim:
Hmm, not obvious to me that this is related to #628. Unless you're certain, @istx25 can you reopen this one?

istx25:
Reopened just in case. I'll do some more digging in the near future.

1ec5:
#714 is one contributing factor to the unexpected behavior reported in this issue.

johnfairh:
Culprit was the preceding header - fixed by #766 in 0.8.0.

Issue #628, opened 2016-08-01
Example Swift code can not contain blank lines


blochberger:
I documented example usage of a function by providing Swift code with

`````` swift
/**
    Does something.

    #### Example
    ```swift
    class SomethingSpecific: Something {
        typealias Error = String

        func doSomething() {
            …
        }
    }
    ```
*/
public class Something {
}
``````

And the output is something like

> Does something.
> 
> **Example**
> 
> ``` swift
> class SomethingSpecific: Something {
>   typealias Error = String
> 
> <span class="kd">func</span> <span class="nf">doSomething</span><span class="p">()</span> <span class="p">{</span>
>    <span class="err">…</span>
> <span class="p">}</span>
> 
> }
> ```

It works fine in the `README.md` file I use as the appendix.

istx25:
Strange! Thanks for filing this issue. I'll do some digging and come back with what I find.

jpsim:
Fixed in #766.

PR #766, opened 2017-04-06
Change markdown header html to stop corruption - issue 628

johnfairh:
Found this on the way to fixing something else...

See issues #628 and #630: a header in source code markdown can cause subsequent markup to be munged.  This happens when Jazzy sends the html 'abstract' generated by SourceKitten.make_doc_info() back through Redcarpet in DocBuilder.render_item().  The problem is that Redcarpet does html passthrough only for block elements. JazzyHTML.header() generates a sequence starting `<a>`.  This is not a block element and so things go poorly for everything up to the next blank line.

I’m pretty sure the `<a>` parts don't do anything and aren't required -- looks like a legacy of importing the github readme format into jazzy many moons ago.  So taking it out fixes these issues and has less html.  All the examples I have/can find look identical.

(I can't follow why jazzy needs to run the html item abstract through Redcarpet.  Maybe some item abstracts can be markdown here?? Insight appreciated.)

If you'd like this change then I will go ahead with the integration-specs part -- wondered if I could take them up to Xcode 8.3 first (different PR maybe?  8.3 results look OK to me: a few changed USRs and it picks up a previously missed typealias in Moya.)

### DIFF
--- a/lib/jazzy/jazzy_markdown.rb
+++ b/lib/jazzy/jazzy_markdown.rb
@@ -13,10 +13,9 @@ module Jazzy
                       .sub(/^-/, '')
                       .sub(/-$/, '')
 
-      "<a href='##{text_slug}' class='anchor' aria-hidden=true>" \
-        '<span class="header-anchor"></span>' \
-      '</a>' \
-      "<h#{header_level} id='#{text_slug}'>#{text}</h#{header_level}>\n"
+      "<h#{header_level} id='#{text_slug}' class='heading'>" \
+        "#{text}" \
+      "</h#{header_level}>\n"
     end
 
     # List from
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -13,7 +13,9 @@
 
 ##### Bug Fixes
 
-* None.
+* Headers in source code markdown no longer cause corruption.  
+  [John Fairhurst](https://github.com/johnfairh)
+  [#628](https://github.com/realm/jazzy/issues/628)
 
 ## 0.7.5
 
--- a/lib/jazzy/doc_builder.rb
+++ b/lib/jazzy/doc_builder.rb
@@ -305,25 +305,21 @@ module Jazzy
       abstract = (item.abstract || '') + (item.discussion || '')
       item_render = {
         name:                       item.name,
-        abstract:                   render_markdown(abstract),
+        abstract:                   abstract,
         declaration:                item.declaration,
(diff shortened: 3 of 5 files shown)


lib/jazzy/jazzy_markdown.rb
```diff
@@ -13,10 +13,9 @@ def header(text, header_level)
                       .sub(/^-/, '')
                       .sub(/-$/, '')
 
-      "<a href='##{text_slug}' class='anchor' aria-hidden=true>" \
-        '<span class="header-anchor"></span>' \
```
1ec5: Is it possible that some themes display this anchor as a link icon when hovering over the header, to make it easier to share a link to a particular section?

johnfairh: It might be tough to do exactly that because the anchor + span do not overlap the header.  You're right though that there could be a user theme that gives this anchor some visual or relies on this html somehow.  I understand if that rules out this approach.

1ec5: Don’t mind me – just pointing out something that came to mind. 😅 I don’t know if any themes actually do this. It would be feasible for a third-party theme to reimplement this functionality in JavaScript anyhow.

johnfairh:
Added a commit to straighten out handling of redcarpet output so that it is never sent back through again.  Only effect on integration_specs output is removal of some blank lines and addition of a few `<p>` tags in the search json -- neither have any effect on search function/how pages look.

jpsim:
It's unfortunate that Circle didn't run tests for this PR, since I'd want to look at the results of the integration specs to confirm this change is ok. I also no longer have Xcode 8.2 installed on my machine so I can't run them locally 😕 .

Can anyone rebuild the integration specs and get this running/passing CI?

johnfairh:
Diff of the specs with this PR using Xcode 8.2 here:
https://github.com/johnfairh/jazzy-integration-specs/tree/jazzy-issue-628-xcode82

(I think Circle has a special toggle for 'enable CI for PRs *from forks*')

jpsim:
Switched it 😄 

image

jpsim:
I went ahead and merged your branch to the integration specs `master`. The next build on `jazzy@master` will fail, then I'll update `jazzy@master` to the latest integration specs.

Thanks!

jpsim:
d2abcc9

johnfairh:
Great - cheers.

Issue #546, opened 2016-04-29
Jazzy requires newlines in Swift doc comments where Swift/Xcode doesn't


kballard:
Jazzy isn't interpreting Swift doc comment directives the same way Xcode does. In particular, Jazzy seems to require a blank line before directives where Xcode doesn't.

For example, with the following declaration:

``` swift
/// The default `HTTPManager` instance.
/// - SeeAlso: `HTTPManagerConfigurable`.
public let HTTP = HTTPManager(shared: true)
```

Xcode handles this properly:

![screen shot 2016-04-29 at 11 25 05 am](https://cloud.githubusercontent.com/assets/714/14925596/0b5d8fb0-0dfd-11e6-95f5-8e05f3d1f953.png)

But Jazzy doesn't:

![screen shot 2016-04-29 at 11 25 23 am](https://cloud.githubusercontent.com/assets/714/14925609/15a10baa-0dfd-11e6-89e9-4e42a2a02451.png)

jazzy version: 0.6.0

kballard:
I also see this occur with `- Parameter` and `- Note` directives, and I assume it occurs with every directive.

jpsim:
Thanks for reporting... this has been brought up before and my reaction then was to close it but now I'm thinking it'd be best to mirror Xcode's semantics.

The reason for this discrepancy is that jazzy (well really [redcarpet](https://github.com/vmg/redcarpet)) requires strict Markdown syntax, whereas Xcode doesn't.

However, I really think the best move here would be to move to using [Apple's fork of CommonMark](https://github.com/apple/swift-cmark) that's used to generate QuickHelp docs to match Xcode's behavior exactly.

kballard:
Sounds like a good idea. I definitely think Jazzy should match the parsing of Xcode, because people writing documentation comments will likely be using Xcode for them and will expect Jazzy to match. Using the actual implementation Xcode uses sounds like a good way to get this behavior.

istx25:
I definitely agree with @kballard on this. Has any progress been made?

jpsim:
There's no doubt about it, if Xcode supports a doc format, we should strive to support it too.

Moving the implementation to CommonMark is still a major project though, and would probably incur some regressions, given that redcarpet supports certain useful non-standard Markdown features that our users rely on, like mixed HTML/Markdown.

johnfairh:
This has been fixed in master for a while now.

PR #782, opened 2017-04-26
Don't require blank lines in markdown before lists

johnfairh:
From at least #546.

Redcarpet has a flag called ‘lax_spacing’ which is documented as being to do with html.  From [the code](https://github.com/vmg/redcarpet/blob/68bfb5b79a08de4ee16e06905f27adb25eabbbe9/ext/redcarpet/markdown.c#L1660), this flag also stops redcarpet from requiring a blank line before lists and backticks-code blocks as well as before html.  The flag has no other effect.

Setting the flag means that redcarpet behaves in the same way as Xcode’s parser and how many humans (including this one) expect things to work.

Specs changed: this change fixed the parsing of a couple of mistakes (missing blank lines) in Siesta and I added a new test file.  Specs PR on the way.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -13,6 +13,10 @@
 
 ##### Bug Fixes
 
+* Blank line no longer needed before lists or code blocks.  
+  [John Fairhurst](https://github.com/johnfairh)
+  [#546](https://github.com/realm/jazzy/issues/546)
+
 * Linking to headers in apple theme gives correct vertical alignment.  
   [John Fairhurst](https://github.com/johnfairh)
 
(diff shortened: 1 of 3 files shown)


jpsim:
This is so very fantastic! 👏 

I'm puzzled as to why [CI passed](https://circleci.com/gh/realm/jazzy/52) even though the submodule wasn't updated to point to realm/jazzy-integration-specs#29.

SDGGiesbrecht:
> This is so very fantastic! 👏

I concur.

johnfairh:
So rare to find an actual "do what I mean" flag 😃 

The unexpected CI pass turns out to be because of a typo in the integration specs that is omitting checking of the Siesta files.  Committed a fix now so CI should fail (!)

jpsim:
Great work @johnfairh 👏

johnfairh:
tyty!

Issue #476, opened 2016-02-05
Jazzy could not recognize the `List-Bullet` tag.


kiary:
Just got this error while generating the doc. The problem is in a comment like this

```
/**
Load accounts that are synchronized in the iPhone

Load the different sources (Facebook, Local, Gmail, Outlook,...) found in the 
Address Book of the iPhone and gives each one it appropriate name.

 - parameter addressBook: The ABAddressBook object relative to the local Address Book

 - returns: A tuple of objects:
    * An array of strings with the names of the accounts
    * An array of ABRecord with the actual sources objects relative to the names
    * A dictionary with the correspondences between accounts names and their ID 
            for future retrieval

 */
```

The list inside the returns tag is not visible in the doc.

jpsim:
Sorry for the delay in getting back to you @kiary, I think the issue here is that jazzy expects the contents of `returns` to be a single markdown statement, not multiple ones, like an unordered list. I suppose this works well in most cases, but certainly not this one. In any case, I'd consider this a bug and something we should fix, but I'd encourage you to work around this for now.

chillok:
@jpsim Also getting this error - though it's not in a return. Will I file a fresh bug? This is essentially the code.

```
/**
     Pushes a specific Screen from the specified UIViewController.
     
     - parameter style:              The design style of the Screen.
                                        
        ### Available Styles:
        - Screen.A
        - Screen.B
     
     - parameter fromViewController: the `UIViewController` to push from.
     */
```

mbmadaj:
I can also confirm bullet list is not recognized by jazzy:

`
public class Test {
    /**
     Does something.

     - note:
     Some note

     - parameter nested: Any?

         Available option:
         * something

    - parameter other: String

     - returns: AnyObject?
     */
    public func xyz(nested:Any?, other: String) -> AnyObject? {
        return nil
    }
}
`

[Test.swift.zip](https://github.com/realm/jazzy/files/227345/Test.swift.zip)

![jazzy parsing result](https://cloud.githubusercontent.com/assets/1888852/14665661/ee2929c4-06d4-11e6-9e40-fd8946d1e238.png)

bolismauro:
This works for me
```
/**
  List:
  
  - B

  - A
```

istx25:
Any progress on this?

jpsim:
Bug still exists.

andyjohns:
[...]

PR #795, opened 2017-05-01
Allow all markdown in param and returns callouts

johnfairh:
From #476 etc.
This suggests how to display all param/returns markdown like Xcode.  It works by snooping the callout rendering code -- least ugly solution I could find :/  Recommend reading commits separately: first is the actual code, second is a tidy of the ruby interface.

Spec changes show whitespace and apostrophe encoding changes galore, none of which affect browser view; one issue fixed in Siesta (`*and*` previously lost); a few new tests; and an issue I’d like feedback on:

Current jazzy gets param text from sourcekit’s *full_as_xml*.  This is smart: it ascends the Swift inheritance tree to find docs.  This has odd results for example [AnyRealmCollection.index(after:)](https://realm.io/docs/swift/2.6.2/api/Classes/AnyRealmCollection.html#/s:FPs14_IndexableBase5indexFT5afterwx5Index_wxS0_).

This method has no docs [in the source code](https://github.com/realm/realm-cocoa/blob/3df0fb6ec06a0f3c5ca49c88bd51ab5b64f4f0af/RealmSwift/RealmCollection.swift#L530).  It gets its body doc via a sourcekitten bug from the previous declaration, and the param/returns doc from Swift.Collection via *full_as_xml*.

Half a dozen instances of this pattern across Realm + Moya.  So not that common.

With this PR as-is, the inherited param/returns docs go away because they are not in the doc comment.  On balance I feel this is more consistent with jazzy generally not displaying inherited docs.  OTOH those docs are correct.  Please say if you’d prefer these to be retained -- easy change.

(Aside: I wonder vaguely about a change of approach to stop parsing the declaration markdown, leaving that to sourcekit, and instead xslt (etc.) the full_as_xml to html. This would overcome all the markdown parser differences and make use of sourcekit’s ‘track down inherited docs’ feature.  Gives a nice pipeline markdown->xml->html.)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,7 +6,9 @@
 
 ##### Enhancements
 
-* None.
+* Allow all markdown in returns and parameter description callouts.  
+  [John Fairhurst](https://github.com/johnfairh)
+  [#476](https://github.com/realm/jazzy/issues/476)
 
 ##### Bug Fixes
 
--- a/lib/jazzy/doc.rb
+++ b/lib/jazzy/doc.rb
@@ -19,7 +19,7 @@ module Jazzy
         "&copy; #{year} [#{config.author_name}](#{config.author_url}). " \
         "All rights reserved. (Last updated: #{date})"
       )
-      Jazzy.copyright_markdown.render(copyright).chomp
+      Markdown.render_copyright(copyright).chomp
     end
 
     def jazzy_version
--- a/lib/jazzy/doc_builder.rb
+++ b/lib/jazzy/doc_builder.rb
@@ -220,7 +220,7 @@ module Jazzy
       doc = Doc.new # Mustache model instance
       name = doc_model.name == 'index' ? source_module.name : doc_model.name
       doc[:name] = name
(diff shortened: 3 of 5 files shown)


jpsim:
This looks great. I used `git diff -w -U0 --word-diff-regex="[^[:space:]]" master` to review https://github.com/realm/jazzy-integration-specs/pull/30 ignoring whitespace-only changes that don't affect the rendered HTML.

The fact that some of the declarations were displaying inherited docs is largely an accident rather than part of a holistic design for supporting inherited docs, so I'm fine with those small removals.

Excellent work here. I'll take care of updating this branch to point to the merged specs changes and getting this to pass CI.

PS: I've granted you write access to the integration specs and the main jazzy repos to make it easier for you to contribute in the future. Please keep up the great work! 👏

jpsim:
Actually it appears I don't have push access to your fork. For now I'll merge manually outside GitHub, but I'd appreciate if in the future you either gave me write access to your fork or made PR branches directly in this repo with your newly granted access rights 😉 . Thanks!

johnfairh:
Thanks for doing the admin work!  Will try not to make a mess in the repos 😛

Issue #636, opened 2016-08-12
error: use of '@import' when modules are disabled


TitouanVanBelle:
I am running into these errors when running the following command

```
15:08:48 - tito@lt0311-2 Jazzy(master)> be jazzy --objc --umbrella-header Jazzy/SDK.h --theme apple --framework-root . --sdk iphoneos --module FyberSDK
/Users/tito/Documents/workspace/fyber/Jazzy/Jazzy/SDK.h:9:1: error: use of '@import' when modules are disabled
/Users/tito/Documents/workspace/fyber/Jazzy/Jazzy/Enum.h:9:1: error: use of '@import' when modules are disabled
/Users/tito/Documents/workspace/fyber/Jazzy/Jazzy/Enum.h:14:36: error: function definition declared 'typedef'
/Users/tito/Documents/workspace/fyber/Jazzy/Jazzy/Enum.h:14:9: warning: type specifier missing, defaults to 'int' [-Wimplicit-int]
/Users/tito/Documents/workspace/fyber/Jazzy/Jazzy/Enum.h:18:5: error: use of undeclared identifier 'MyEnumA'
/Users/tito/Documents/workspace/fyber/Jazzy/Jazzy/Enum.h:23:5: error: use of undeclared identifier 'MyEnumB'
/Users/tito/Documents/workspace/fyber/Jazzy/Jazzy/Enum.h:28:5: error: use of undeclared identifier 'MyEnumC'
building site
jam out ♪♫ to your fresh new docs in `docs`
```

I checked and modules are enabled for the target

```
//:configuration = Debug
CLANG_ALLOW_NON_MODULAR_INCLUDES_IN_FRAMEWORK_MODULES = NO
CLANG_ENABLE_MODULES = YES

//:configuration = Release
CLANG_ALLOW_NON_MODULAR_INCLUDES_IN_FRAMEWORK_MODULES = NO
CLANG_ENABLE_MODULES = YES

//:completeSettings = some
CLANG_ALLOW_NON_MODULAR_INCLUDES_IN_FRAMEWORK_MODULES
CLANG_ENABLE_MODULES
CLANG_MODULES_AUTOLINK
```

I am using the following version of jazzy

```
15:08:46 - tito@lt0311-2 Jazzy(master)> be jazzy -v
jazzy version: 0.7.0
```

I created a simple project that can reproduce the issue
https://github.com/TitouanVanBelle/JazzySampleProject

istx25:
[...]

PR #797, opened 2017-05-02
Fix an issue when additional xcodebuild arguments are ignored in Objective-C mode

mgrebenets:
When running `jazzy` in Objective-C mode like this:

```bash
jazzy --objc \
  --sdk iphonesimulator \
  --umbrella-header Sources/MyFramework.h \
  --module MyFramework \
  --framework-root Sources \
  --no-download-badge \
  --module-version 1.0.0 \
  --output objc-docs
```

`jazzy` adds the following arguments for `sourcekitten` under the hood:

```ruby
["doc",
  "--objc", 
  "/Users/grebenma/Projects/myframework.ios/Sources/MyFramework.h", 
  "--", "-x", "objective-c", 
  "-isysroot", "/Applications/Xcode.app/Contents/Developer/Platforms/iPhoneSimulator.platform/Developer/SDKs/iPhoneSimulator10.2.sdk",
   "-I", "/Users/grebenma/Projects/myframework.ios/Sources"
]
```

But then I want to add `-fmodules` argument for compiler command using `-x` option:

```bash
jazzy --objc \
  --sdk iphonesimulator \
  --umbrella-header Sources/MyFramework.h \
  --module MyFramework \
  --framework-root Sources \
  --no-download-badge \
  --module-version 1.0.0 \
  --output objc-docs \
  -x -fmodules
```

Now `jazzy` will not autogenerate all those build options for `sourcekitten` and the final arguments list will be:

```ruby
["doc", "--", "-fmodules"]
```

This change will make sure that user-specified arguments are _appended_ to automatically generated arguments.

This helps with issues like https://github.com/realm/jazzy/issues/636.

One caveat that now in Objective-C mode users should not use `--` together with `-x` because `--` is already added by `jazzy`.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,7 +2,9 @@
 
 ##### Breaking
 
-* None.
+* Fix an issue when additional xcodebuild arguments are ignored in Objective-C mode.
+  [Maksym Grebenets](https://github.com/mgrebenets)
+  [#797](https://github.com/realm/jazzy/issues/797)
 
 ##### Enhancements
 
--- a/lib/jazzy/sourcekitten.rb
+++ b/lib/jazzy/sourcekitten.rb
@@ -185,13 +185,10 @@ module Jazzy
     end
 
     def self.objc_arguments_from_options(options)
-      arguments = []
-      if options.xcodebuild_arguments.empty?
-        arguments += ['--objc', options.umbrella_header.to_s, '--', '-x',
-                      'objective-c', '-isysroot',
-                      `xcrun --show-sdk-path --sdk #{options.sdk}`.chomp,
-                      '-I', options.framework_root.to_s]
-      end
+      arguments = ['--objc', options.umbrella_header.to_s, '--', '-x',
+                   'objective-c', '-isysroot',
+                   `xcrun --show-sdk-path --sdk #{options.sdk}`.chomp,
+                   '-I', options.framework_root.to_s]
       # add additional -I arguments for each subdirectory of framework_root
       unless options.framework_root.nil?
         rec_path(Pathname.new(options.framework_root.to_s)).collect do |child|


jpsim:
Could you please add a changelog entry? This is technically a breaking change.

As an aside, I've never liked Jazzy's ObjC interface. Making the `xcodebuild` argument extraction work the same way it does for Swift will eternally be on my list of things to do. See jpsim/SourceKitten#88.

jpsim:
I'm not sure if this is the best fix, to be honest. This is fine if you only want to _append_ to jazzy's autogenerated arguments, but if you want full control over them, you lose that with this PR.

Also, not only is your changelog entry missing the two trailing spaces, but you've stripped them from all existing changelog entries in 4241bb4. Please refer to [CONTRIBUTING.md](https://github.com/realm/jazzy/blob/master/CONTRIBUTING.md#tracking-changes) for formatting guidelines for our changelog. Thanks.

mgrebenets:
@jpsim 

True, but then at current state as a user I don't have much control either.
As soon as I add at least one `-x ...` option, all the options I passed using `--objc`, `--umbrella-header`, `--module` and so on are now silently ignored. To make ObjC work I pretty much have to duplicate them all in `-x ...` part, including the logic for recursively finding all subdirectories to pass with `-I` option, the part where `objective-c`, `-sysroot` and other things are added.

Maybe the other way to approach this is to add `-fmodules` in `self.objc_arguments_from_options`?

This would work under assumption that the project in question is using modules.
Should be pretty safe assumption, since modules have been introduces quite a long time ago and _probably_ by this time only few legacy projects would be built without use of Clang modules...


Missed the spaces guideline 🙃, have been stripping them by default in the editor.

mgrebenets:
Here's an attempt to approach this problem from another angle: https://github.com/realm/jazzy/pull/798

mgrebenets:
Closing in favor of https://github.com/realm/jazzy/pull/798

PR #798, opened 2017-05-03
Add -fmodules flag for default objective-c options list

mgrebenets:
Another take on #636.
The assumption is that recent projects would have clang modules enabled by this time.

If a project still doesn't use modules, then probably it's build with Xcode 6 or alike and could use `appledoc`...

If user does have a project which doesn't use modules, then probably they should provide all required options via `-x`. Default configuration will work out of the box for most popular configurations.

This is less breaking than #797


jpsim:
Thanks!

Issue #801, opened 2017-05-05
undefined method `render' for Redcarpet::Render::SmartyPants:Module


groue:
Hello,

I've just installed jazzy 0.8.1, and it can't generate documentation any longer:

```
$ jazzy \
  --clean \
  --author 'Gwendal Roué' \
  --author_url https://github.com/groue \
  --github_url https://github.com/groue/GRDB.swift \
  --github-file-prefix https://github.com/groue/GRDB.swift/tree/v0.107.0 \
  --module-version 0.107.0 \
  --module GRDB \
  --root-url http://groue.github.io/GRDB.swift/docs/0.107.0/ \
  --output Documentation/Reference \
  --podspec GRDB.swift.podspec
Analyzing dependencies
Fetching podspec for `GRDB.swift` from `/Users/groue/Documents/git/groue/GRDB.swift-master.swift`
...
/Users/groue/.rbenv/versions/2.2.2/lib/ruby/gems/2.2.0/gems/jazzy-0.8.1/lib/jazzy/jazzy_markdown.rb:151:in `render_param_returns': undefined method `render' for Redcarpet::Render::SmartyPants:Module (NoMethodError)
	from /Users/groue/.rbenv/versions/2.2.2/lib/ruby/gems/2.2.0/gems/jazzy-0.8.1/lib/jazzy/jazzy_markdown.rb:143:in `list_item'
	from /Users/groue/.rbenv/versions/2.2.2/lib/ruby/gems/2.2.0/gems/jazzy-0.8.1/lib/jazzy/jazzy_markdown.rb:165:in `render'
	from /Users/groue/.rbenv/versions/2.2.2/lib/ruby/gems/2.2.0/gems/jazzy-0.8.1/lib/jazzy/jazzy_markdown.rb:165:in `render'
	from /Users/groue/.rbenv/versions/2.2.2/lib/ruby/gems/2.2.0/gems/jazzy-0.8.1/lib/jazzy/sourcekitten.rb:316:in `make_doc_info'
	from /Users/groue/.rbenv/versions/2.2.2/lib/ruby/gems/2.2.0/gems/jazzy-0.8.1/lib/jazzy/sourcekitten.rb:382:in `block in make_source_declarations'
	from /Users/groue/.rbenv/versions/2.2.2/lib/ruby/gems/2.2.0/gems/jazzy-0.8.1/lib/jazzy/sourcekitten.rb:343:in `each'
[...]

PR #802, opened 2017-05-05
Call smartypants differently to avoid ruby 2.2 error

johnfairh:
See #801, some incompatibility between code added in 3c16e0d and ruby 2.2.
Don't understand why the problem happens on 2.2 and not on 2.0/2.4.
This change calls the same routine in a different way - no effect on html.
Works on 2.0/2.2.2/2.4.1.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,7 +10,9 @@
 
 ##### Bug Fixes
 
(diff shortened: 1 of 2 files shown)


jpsim:
Thanks for fixing this!

PR #804, opened 2017-05-05
Fix space before parameter abstract in output

nrbrook:
It seems that the content passed to Markdown contains `codepoint U+001F INFORMATION SEPARATOR ONE` and a newline at the start, causing a newline (and therefore space when displayed) before the first character in the html output. This strips it.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,8 +9,9 @@
 * None.
 
 ##### Bug Fixes
-
-* None.
+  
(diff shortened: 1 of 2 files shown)


johnfairh:
Hi there --- the CI results show that the test projects all came out with identical byte-for-byte HTML before + after this change.  So I'm not sure yet that jazzy itself is introducing these characters.  Could they be present in the source code of the target project?  Are you able to share an example of the project + the output?

jpsim:
Closing due to inactivity.

PR #815, opened 2017-05-12
Update SourceKittenFramework to 0.17.3

jpsim:


jpsim:
Changes to integration specs were reviewed in jpsim/SourceKitten#375.

Issue #620, opened 2016-07-23
some extensions render as undocumented


aramikg:
I have a few extension that are being generated just fine but then also some that are being generated as undocumented.

Undocumented extensions:
- UIBarItem
- UIBarButtonItem
- UIViewController

sourcekitten 0.13 and 0.7 version of jazzy

seanoshea:
I'm running into this exact same issue with http://cocoadocs.org/docsets/FuzzyMatchingSwift/0.1.0/Extensions/_ArrayType.html. Source code is available at https://github.com/seanoshea/FuzzyMatchingSwift/blob/develop/FuzzyMatchingSwift/Classes/FuzzyMatching.swift#L37 using the `0.7.0` version of `jazzy`

istx25:
Hi @aramikg and @seanoshea. Thanks for submitting this issue (and adding some additional context, @seanoshea). I'll review all of this and see what we can do.

jpsim:
Fixed in #819

PR #819, opened 2017-05-16
Display doc comments for all swift extensions

johnfairh:
See #454, #620.
Now sourcekitten is passing on doc comments of extensions, this PR displays them.  The corner case where this wasn't working is Swift extensions of Objective C classes like `Foundation.NSDate`: in this case sourcekit does not provide the full_as_xml field.

Specs changes show one fix in Siesta demonstrating exactly this, and another 'fix' in realm-swift where a previously missed doc comment is now displayed --- [the declaration](https://realm.io/docs/swift/2.7.0/api/Classes/Object.html#/) shouldn't be present at all, I think, but that's another story + not a regression here...

edit with latest push: also propose dropping the 'undocumented' label on swift extensions of objc classes *without* doc comments, for consistency with other extensions + because they do not contribute to undocumented%.  Spec changes reflect that.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,7 +10,9 @@
 
 ##### Bug Fixes
 
-* None.
+* Fix missing doc comments on some extensions.  
+  [John Fairhurst](https://github.com/johnfairh)
+  [#454](https://github.com/realm/jazzy/issues/454)
 
 ## 0.8.2
 
--- a/lib/jazzy/sourcekitten.rb
+++ b/lib/jazzy/sourcekitten.rb
@@ -281,8 +281,15 @@ module Jazzy
       if objc || should_mark_undocumented(doc['key.kind'], filepath)
         @stats.add_undocumented(declaration)
       end
+
       return nil if !documented_child?(doc) && @skip_undocumented
+
       make_default_doc_info(declaration)
+      if doc.key?('key.doc.comment')
+        # eg. swift extension of a type from a non-swift module
+        declaration.abstract = Markdown.render(doc['key.doc.comment'])
+      end
+      declaration
     end
 
     def self.parameters(doc, discovered)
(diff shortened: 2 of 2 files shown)


SDGGiesbrecht:
What happens when no documentation is provided?

Will this introduce new entries in `undocumented.json`, or will it just remove the peculiar “Undocumented” note from the generated documentation?

It seems odd to me to require documentation when the symbol is already documented at its declaration (even if in another module), and the extension documentation is inaccessible to Quick Help and friends.

On the other hand, thumbs up if this stops Jazzy from publishing the nonsense text “Undocumented” under each extended type. That has been annoying me for a while.

I don’t mind if I will have to add `:nodoc:` or something similar to extensions to suppress warnings, but I’m concerned that it be possible to do in a way that does not make the symbol itself vanish from the documentation—especially if its descendants disappear with it.

johnfairh:
If no doc were provided then you would get the 'nonsense' text 😄 as today, no change.

The only thing this change does is display the doc comment -- if you wrote one -- instead of 'Undocumented'.  I think it is often natural + useful to write doc comments on extensions of existing types to explain their purpose in the context of a module's complete documentation; several user issues along this line.

So, you would be able to get rid of 'undocumented' with an empty doc comment '///'.

Extensions of types from other modules never contribute to `undocumented.json`.

...so it is a bit unhelpful to stick 'undocumented' on them without warning.
And some extensions have no need of documentation because covered in content docs / would just be 'Some useful methods on Foo'. (plus your point wrt Xcode).
And extensions of Swift types (local or foreign module) do not behave like this.

OK, I'm convinced: we should never see this particular 'Undocumented'.

New proposal for undocumented symbols:
1. Display 'Undocumented.' for the abstract if and only if the symbol goes in `undocumented.json` and so affects documented%: we want the user to write some docs.
2. Otherwise display the doc comment as the abstract, which may be blank: we don't mind if the user doesn't write docs, but if they do then we'll display them.

Thanks a lot for the comments.

SDGGiesbrecht:
Sounds perfect, @johnfairh. Thanks for listening.

jpsim:
Awesome, thanks @johnfairh 👏

johnfairh:
cheers!

Issue #732, opened 2017-01-12
Extensions to Core Data managed classes in Xcode 8/Swift3/iOS10 are skipped if "--skip-undocumented" is used


esopher:
When using Core Data in Xcode 8/Swift 3/iOS 10, the class files for entities no longer need to be generated, but when extending an entity class (NSManagedObject), (and documenting it), Jazzy skips the extension if "--skip-undocumented" is used. Presumably this is because the parent class is not documented, and in fact doesn't exist, at least not in a .swift file in the project?

To reproduce this, create a new single-view application for iOS using Core Data, add an Entity named "Entity" to the model, and add an Entity.swift file to the project that extends the "Entity" class:

```
/// An extension to Entity.
extension Entity {
    /// A great function!
    func aGreatFunction() {
        print("It does nothing!")
    }
}
```
Run "jazzy --skip-undocumented" in the project directory and see that the extension documentation is not included.

This  might be a separate issue, but I just realized that when not using "--skip-undocumented", the extension is documented but the documentation is added to the base entity class, which is still undocumented, and listed under Classes instead of Extensions.

jpsim:
Fixed in #825.

johnfairh:
The first attempt at 825 did fix this particular case (extension + type in same module) -- the eventual one did not -- will get there eventually though.

jpsim:
Ah, sorry for the misunderstanding

PR #508, opened 2016-04-03
Support documenting "substructures" such as Extensions.

freak4pc:
This should resolve issue #502 where some documents didn't get recognised as being documented. 

This is due to the fact some more complex parts of code/documentation don't have a `key.doc.full_as_xml` property so they're being discarded as "undocumented", even though some of these have a `key.substructure` property which provides additional documentation information that should be considered.

Issue #502, opened 2016-03-27
Docs for NSError extension (specifically) don't get parsed. 


freak4pc:
I have some extensions that are all being parsed correctly, for example, this gets detected with no issues and generates valid HTML docs: 

``` swift
public extension String {
    //MARK: - String Extras

    /// A capitalized address string.
    /// e.g. `5TH AVENUE ST` returns `5th Avenue St`
    public var capitalizedAddress: String {
```

But specifically my NSError extensions are being skipped for some reason and docs are never generated even though they follow the same format: 

``` swift
public extension NSError {
    //MARK: - NSError Extras

    /// Return a `THResponse` object corresponding to the current `NSError`.
    public var asResponse: THResponse {
    ...
    ...
```

This is how it looks: 
![image](https://cloud.githubusercontent.com/assets/605076/14064793/d9477cd2-f418-11e5-80e3-ff741668cacc.png)

This is even weirder since jazzy outputs parsing of the file, but doesn't generate the docs from it e.g. `Parsing NSError.Exts.swift (27/47)` . 

Would appreciate any help on this !

Kind Regards,
Shai.

freak4pc:
This seems to be the case for UIColor, UIFont, CLLocation extensions as well. But works for Float, String, UIView extensions. 

I thought this might be related to Swift types vs. Obj-C types but UIView works fine ...

segiddins:
Do you have the necessary import statements in those files?

radutzan:
This is happening to me with a UIFont extension. The file has `import UIKit` at the top. A different file with  a UIImage extension doesn't have the issue. I tried moving the UIFont extension to the same file where the UIImage extension is, but no luck.

freak4pc:
@segiddin what do you mean by that? Those files have the regular imports just like the rest of my files that are recognized. This seems specific to extensions for those classes.

segiddins:
In that case, a project that reproduces the issue would be the first step towards debugging.

freak4pc:
Absolutely. Taking care of that now.

freak4pc:
I seem to have tracked down the bug. 
It doesn't create documentation for that file when I execute with the `--skip-undocumented`. 

e.g. this would generate documentation for that file 
`jazzy --swift-version 2.2 --min-acl internal`
![image](https://cloud.githubusercontent.com/assets/605076/14108452/2a412e0e-f5c6-11e5-9d4a-a75c8cc14af9.png)

this won't, even though that file IS documented : 
[...]

PR #825, opened 2017-05-21
Make --skip-undocumented less aggressive

johnfairh:
Continuing @freak4pc’s investigation from #508, also issue #732.
This PR fixes `--skip-undocumented` so that an undocumented declaration is not skipped if any of its subtree is documented.

Spec changes show a new extension to `URLRequest` being discovered in AlamoFire (which sets this option) -- affects left nav so all files show changes.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -26,9 +26,9 @@
   [#824](https://github.com/realm/jazzy/issues/824)
 
 * Stop --skip-undocumented from skipping documented items nested
-  inside undocumented items.  
+  inside extensions of types from other modules.  
   [John Fairhurst](https://github.com/johnfairh)
-  [#732](https://github.com/realm/jazzy/issues/732)
+  [#502](https://github.com/realm/jazzy/issues/502)
 
 ## 0.8.2
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -25,6 +25,11 @@
   [JP Simard](https://github.com/jpsim)
   [#824](https://github.com/realm/jazzy/issues/824)
 
+* Stop --skip-undocumented from skipping documented items nested
+  inside undocumented items.  
+  [John Fairhurst](https://github.com/johnfairh)
+  [#732](https://github.com/realm/jazzy/issues/732)
+
 ## 0.8.2
 
 ##### Breaking
--- a/Rakefile
+++ b/Rakefile
@@ -55,23 +55,14 @@ begin
     # Remove files not used for the comparison
     # To keep the git diff clean
     files_glob = 'spec/integration_specs/*/after/{*,.*}'
-    exec_output_glob = 'spec/integration_specs/*/after/execution_output.txt'
     files_to_delete = FileList[files_glob]
       .exclude('**/.', '**/..')
       .exclude('spec/integration_specs/*/after/*docs',
-               exec_output_glob)
+               'spec/integration_specs/*/after/execution_output.txt')
       .include('**/*.dsidx')
       .include('**/*.tgz')
     files_to_delete.each do |file_to_delete|
       sh "rm -rf '#{file_to_delete}'"
-    end
-
-    # Remove version numbers from CocoaPods dependencies
-    # to make specs resilient against dependecy updates.
-    FileList[exec_output_glob].each do |file|
-      sanitized = File.open(file, &:read)
-                      .gsub(/(Installing \w+ )\((.*)\)/, '\1(X.Y.Z)')
-      File.write(file, sanitized)
     end
 
     puts
--- a/Rakefile
+++ b/Rakefile
@@ -55,14 +55,23 @@ begin
     # Remove files not used for the comparison
     # To keep the git diff clean
     files_glob = 'spec/integration_specs/*/after/{*,.*}'
+    exec_output_glob = 'spec/integration_specs/*/after/execution_output.txt'
     files_to_delete = FileList[files_glob]
       .exclude('**/.', '**/..')
       .exclude('spec/integration_specs/*/after/*docs',
-               'spec/integration_specs/*/after/execution_output.txt')
+               exec_output_glob)
       .include('**/*.dsidx')
       .include('**/*.tgz')
     files_to_delete.each do |file_to_delete|
       sh "rm -rf '#{file_to_delete}'"
+    end
+
+    # Remove version numbers from CocoaPods dependencies
+    # to make specs resilient against dependecy updates.
+    FileList[exec_output_glob].each do |file|
+      sanitized = File.open(file, &:read)
+                      .gsub(/(Installing \w+ )\((.*)\)/, '\1(X.Y.Z)')
+      File.write(file, sanitized)
     end
 
     puts
--- a/lib/jazzy/sourcekitten.rb
+++ b/lib/jazzy/sourcekitten.rb
@@ -229,7 +229,10 @@ module Jazzy
 
     def self.documented_child?(doc)
       return false unless doc['key.substructure']
-      doc['key.substructure'].any? { |child| documented_child?(child) }
+      doc['key.substructure'].any? do |child|
+        should_document?(child) &&
+          (can_document?(child) || documented_child?(child))
+      end
     end
 
     def self.availability_attribute?(doc)
@@ -302,12 +305,16 @@ module Jazzy
       end
     end
 
+    def self.can_document?(doc)
+      doc.key?('key.doc.full_as_xml')
+    end
+
     # rubocop:disable Metrics/CyclomaticComplexity
     # rubocop:disable Metrics/PerceivedComplexity
     def self.make_doc_info(doc, declaration)
       return unless should_document?(doc)
 
-      unless doc['key.doc.full_as_xml']
+      unless can_document?(doc)
         return process_undocumented_token(doc, declaration)
       end
 
--- a/lib/jazzy/sourcekitten.rb
+++ b/lib/jazzy/sourcekitten.rb
@@ -227,14 +227,6 @@ module Jazzy
       declaration.children = []
     end
 
-    def self.documented_child?(doc)
-      return false unless doc['key.substructure']
-      doc['key.substructure'].any? do |child|
-        should_document?(child) &&
-          (can_document?(child) || documented_child?(child))
-      end
-    end
-
     def self.availability_attribute?(doc)
       return false unless doc['key.attributes']
       !doc['key.attributes'].select do |attribute|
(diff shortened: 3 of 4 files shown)


freak4pc:
Looks good, I'll test this on my codebase a bit later. Thanks ! @johnfairh

jpsim:
> Just realized i'm not a collaborator anymore so can't review this 🤔 weird

Fiiiiiiixed.

jpsim:
One thing I want to make sure of is that `A` or its sub-declarations remain undocumented in the following situations:

```swift
/// :nodoc:
public struct A {
  /// some docs
  public let a: Int
}
```

And:

```swift
public struct A { // < no documentation comment
  /// some docs
  public let a: Int
}
```

johnfairh:
Interesting!

The `:nodoc:` case behaves like that, encountering that immediately stops traversing the tree and does not include `A` in the docs.

The second case though includes `A`, calls it `Undocumented` in order to include the `a` field.  Any members of `A` without doc comments are skipped.  This was my best guess at what [`documented_child`](https://github.com/realm/jazzy/blob/84357d45ed2fbb83c04f25f6128bec9ee54e7651/lib/jazzy/sourcekitten.rb#L230) was intended to do (otherwise why bother with the routine / any kind of recursion).

Looking at the user issues again though I see they are entirely about extensions.  So maybe the way you see this working is, with `--skip-undocumented`:
* As soon as we find an undocumented declaration, skip it + all its children regardless of their doc comments, so there are no `Undocumented` symbols;
* But make sure extensions with doc comments count as 'documented'.

Sound right?

jpsim:
> As soon as we find an undocumented declaration, skip it + all its children regardless of their doc comments, so there are no `Undocumented` symbols;

👍 

> But make sure extensions ~~with doc comments~~ that do not say `Undocumented` count as 'documented'.

I don't think that's what I'd expect. The doc comment in an extension should meaningfully count as documentation for that extension:

```swift
/// Extends `NSString` with functionality useful to this library
extension NSString {
  /// Some insightful docs into what `asdf` does.
  public func asdf() {}
}
```

I think that ideally `--skip-undocumented` would still generate docs for the `NSString` extension with an "Undocumented" description, as long as it has a `documented_child`. IMHO there's no point in adding that to the generated HTML if `--skip-undocumented` and it has no documented children. However, I'd still say that the extension should be included in the `undocumented` count.

Does that make sense?

SDGGiesbrecht:
> The doc comment in an extension should meaningfully count as documentation for that extension

That seems counterintuitive to me, @jpsim. I follow @johnfairh’s logic more easily.

***

I was never even sure Swift itself intended to support markup for `extension` at all. I went to the [official documentation](https://developer.apple.com/library/content/documentation/Xcode/Reference/xcode_markup_formatting_ref/#//apple_ref/doc/uid/TP40016497-CH2-SW1) looking for an answer.

The word used consistently throughout the documentation is “symbol”.

> Use markup [...] to show Quick Help for your Swift code symbols.

The rough understanding in my mind, having never looked it up, was that a computing symbol is a source code entity that is *referenceable* and *unique* (within its scope).

Searching for “symbol (computing)” at Wikipedia redirects to “[Symbol Table](https://en.wikipedia.org/wiki/Symbol_table)” where it says:

> [A] symbol table is a data structure [...] where each identifier (a.k.a. symbol) in a program’s source code is associated with information relating to its declaration [...] in the source.

Following the “[identifier](https://en.wikipedia.org/wiki/Identifier)” link to its page yields:

> An identifier is a name that identifies [...] a unique object.

I’m not sure if the writers of the documentation were concerned with this level of specificity, but by that definition, extensions are not symbols. They have no names and there is no way to reference them, query them, store them in variables, tell them apart, or even tell if the symbols they contain are part of an extension or not. Most significantly, they do not have a unique declaration—that is there purpose, after all, to add functionality to an existing symbol without declaring any new entity.

Normally documentation belongs at the declaration (because it is unique and easily discoverable), not the usage sites. This example is a good illustration:

```swift
/// A variable.
public var variable: String?

func setVariable(to value: String) {

    /// The variable can be set. (Isn’t this documentation misplaced?)
    variable = value
}
```

By the same logic, documentation would belong at the type’s declaration, not its extensions.

```swift
/// A type.
public struct Type { ... }

/// Types can also do things. (This seems misplaced to me as well.)
extension Type { ... }
```

I understand that when a symbol is imported from another module and its declaration is out of the developer’s control, it can be nice to have a fallback location to note any additions, so I agree with using documentation where the user has provided it on an extension (even if Swift never does). However, considering extensions as things that should always be documented leads to absurd situations:

```swift
/// Some useful extensions.
extension ExternallyDefinedClass { ... }

/// Some more useful extensions.
extension ExternallyDefinedClass { ... }

extension ExternallyDefinedClass { ... }

extension ExternallyDefinedClass { ... }

/// Fancy stuff under certain conditions.
extension ExternallyDefinedClass where Component : Equatable { ... }

extension ExternallyDefinedClass where Component : Comparable { ... }
```

In this example:
- Is the developer doing something wrong?
- Which documentation comment should Jazzy output?
  - Should they be combined? How? In what order, especially if they are in separate files?
  - Should they be kept separate from one another? How will the reader find what they are looking for, especially when things that conceptually belong together have to be split up because of constraints?
- How many “undocumented” warnings should be produced?

I like the “undocumented” warnings when they tell me that there is something exposed to users that they cannot look up. I do not like the idea of “undocumented” warnings for distinctions that do not actually exist for the user (such as the differences between various extensions and the corresponding declaration).

johnfairh:
@jpsim yep on the same page, my second bullet was unclear even/especially after edit.  Will revise code.

I'd like to leave any 'change the rules about requiring docs on extension declarations' out of this PR -- @SDGGiesbrecht these are some good questions, the middle ones apply today and are most important to the end user (reader), although I think relatively rare in practice.

SDGGiesbrecht:
Um... when I first posted, I thought this conversation was in #819, where @johnfairh and I had already been talking about something similar, and I feared this was a reversal of decisions made there. I just realized this is a different pull request. (Silly me.)

When I look closer I see that #819 has already been merged. My new understanding (after reading all the referenced issues) is that this pull request is really about bringing `--skip-undocumented` in line with #819. I’m all for that.

I was going to ask if the two (this and #819) could share logic in order to stay in lockstep in the future, but since you, @johnfairh, wrote both, I assume you already know how they interact. (I mention it for only for the sake of anyone else that is reading.)

> I'd like to leave any 'change the rules about requiring docs on extension declarations' out of this PR

Sounds good.

SDGGiesbrecht:
Here is a summary what my expectations would be:

```swift
// ••••••• ••••••• ••••••• ••••••• ••••••• ••••••• •••••••
// Local Types (Structures, Classes, Enumerations, etc.)
// ••••••• ••••••• ••••••• ••••••• ••••••• ••••••• •••••••

/// Documentation (→ Display)
public struct LocalA {
    
    /// Documentation (→ Display)
    public func a() {}

    public func b() {} // No documentation → Skip (& warn)
}

public struct LocalB { // No documentation → Skip (& warn)
    
    /// Documentation (→ But undocumented parent → Skip)
    public func a() {}

    public func b() {} // No documentation → Skip (& warn)
}

// ••••••• ••••••• ••••••• ••••••• ••••••• ••••••• •••••••
// Extensions of Local Types
// ••••••• ••••••• ••••••• ••••••• ••••••• ••••••• •••••••

/// Documentation (→ But declaration already has documentation → Merge (strategy undefined))
extension A /* Same module */ {
    
    /// Documentation (→ Display (merged into declaration method list))
    public func c() {}

    public func d() {} // No documentation → Skip (& warn)
}

extension A /* Same module */ { // No documentation → Parse children (no warning)
    
    /// Documentation (→ Display (merged into declaration method list))
    public func e() {}

    public func f() {} // No documentation → Skip (& warn)
}

/// Documentation (→ But undocumented at declaration → Merge (strategy undefined))
extension B /* Same module */ {
    
    /// Documentation (→ Display (merged into declaration method list))
    public func c() {}

    public func d() {} // No documentation → Skip (& warn)
}

extension B /* Same module */ { // No documentation → Parse children (no warning)
    
    /// Documentation (→ Display (merged into declaration method list))
    public func e() {}

    public func f() {} // No documentation → Skip (& warn)
}

// ••••••• ••••••• ••••••• ••••••• ••••••• ••••••• •••••••
// Extensions of External Types
// ••••••• ••••••• ••••••• ••••••• ••••••• ••••••• •••••••

/// Documentation (→ Display (merge strategy for multiple extensions undefined))
extension C /* Different module */ {
    
    /// Documentation (→ Display)
    public func c() {}

    public func d() {} // No documentation → Skip (& warn)
}

extension C /* Different module */ { // No documentation → Display (without “Undocumented” description) → Parse children (no warning)
    
    /// Documentation (→ Display)
    public func e() {}

    public func f() {} // No documentation → Skip (& warn)
}
```

SDGGiesbrecht:
I updated my last comment to fix some typos and oversights. I mention it here, because I doubt notifications contain the fixes. If you are reading the original in a notification, some details were not what I intended.

SDGGiesbrecht:
Also, not all of what I suggested above should necessarily be solved with in this pull request. As long as this pull request makes steps towards that ideal and not farther from it, then the I support the pull request as is.

johnfairh:
Pushed commit to wind this back a little: PR now is for extensions of external types like `UIView` in `--skip-undocumented` mode to display documented children, fixing problems reported in #502.  As part of this, extensions that end up not containing anything are dropped from docs.

(one note on examples append above: Core Data default codegen is the example real-world use case in my mind for documenting the `B` extensions: user has no control over the class decl + the codegen does not supply a doc comment there.  (For extra-super clarity, PR does not address this.))

SDGGiesbrecht:
> Core Data default codegen is the example real-world use case in my mind for documenting the B extensions: user has no control over the class decl + the codegen does not supply a doc comment there. 

Good catch, I’ll edit the list to note it in place.

SDGGiesbrecht:
Would it be worth adding those examples to the test suite?

Each one could be documented with either:

```swift
/// This should be visible, because...
```

or

```swift
/// This should be hidden, because...
```

Then whenever one appears or disappears in the output diffs, it is easy to see whether the change is wanted or a regression. It would also have the complete rationale right there in the diffs so it would be harder to overlook something when future changes or decisions are made.

Actually, I just convinced myself. I will go ahead and do that. If you think it is a bad idea or what to discuss the details, you can do so in the coming pull request.

SDGGiesbrecht:
You have my support for this already as‐is. I like the results when I test it. Thank you for working on it, @johnfairh.

jpsim:
Sorry for the delay, this is a lovely improvement. 👏 

I'll merge realm/jazzy-integration-specs#32, update this PR to point to the merged commit, re-run this on CI to make sure it still passes, will fix it up if something broke in the months since this was done and get merged.

johnfairh:
No worries - thanks.  Big 👍 on the spec differ change.

PR #831, opened 2017-05-26
Add Test Fixtures for Treatment of Undocumented Symbols

SDGGiesbrecht:
This enables the tests in realm/jazzy-integration-specs#33. A detailed description can be found there.

This does not affect Jazzy itself in any way.


SDGGiesbrecht:
Well, never mind the demonstration then. I thought #825 would register differences, but it seems to have no effect in its current state.

SDGGiesbrecht:
Closing along with realm/jazzy-integration-specs#33. See the [note there](https://github.com/realm/jazzy-integration-specs/pull/33#issuecomment-305683735).

PR #847, opened 2017-07-06
update SourceKitten to latest master

### LINKED ISSUE #333: Members added to extension of nested type show up in the parent

jpsim:
Mostly just pulls in jpsim/SourceKitten#388

Fixes #333.

Definitely a huge improvement, but extensions of nested types picked up from this new codepath aren't handled like other extensions of nested types:

image

Something we might want to align eventually, but not worth holding this up since the current behavior on `master` is just terrible.

Thanks for your hard work on this @johnfairh

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -25,10 +25,14 @@
   [JP Simard](https://github.com/jpsim)
   [#824](https://github.com/realm/jazzy/issues/824)
 
-* Stop --skip-undocumented from skipping documented items nested
+* Stop `--skip-undocumented` from skipping documented items nested
   inside extensions of types from other modules.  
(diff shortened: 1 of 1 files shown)


johnfairh:
Yep - I remember writing something to handle that, must be in a branch/stash somewhere, will dig out tomorrow.

PR #848, opened 2017-07-07
Merge nested type extensions into original types

johnfairh:
This tidies up extensions of nested types so that the extensions are merged into the original
type, just like extensions of non-nested types.

Spec changes in alamofire + realm-swift to reflect this, left-nav changes so all files change :/
Added some more tests for this kind of thing to misc_jazzy_features.

I decided to leave the changelog alone 'cos this is more of #333.


jpsim:
Nice! Thanks so much.

Issue #805, opened 2017-05-10
Auto-linking HTML-escaped operators


reitzig:
Say I have code like this:

```swift
indirect enum Expr {
	case .number(Int)

	/**
		- Seealso: `Expr.&(...)`
	*/
	case .and(Expr, Expr)

	static func &(lhs: Expr, rhs: Expr) -> Expr {
		return .and(lhs, rhs)
	}
}
```

How do I refer to `&` in a way that auto-links? 

Neither `&` (locally), `Expr.&`, `Expr.&(...)`, nor `Expr.&(lhs:rhs:)` work.

johnfairh:
The most reliable way is, in this case, `+(_:_:) or `Expr.+(_:_:)` -- the part after the period is the name that jazzy thinks the symbol has, as show in its doc page.

I also found that `Expr.+(...)` worked for me, the ... works as a sort of wildcard.

reitzig:
Hm... did you also try it in the callout? I'm on v0.8.1, if that's relevant. 

I'm looking at the generated doc right now, and there's no link. (Things are called by different names, but that shouldn't be a problem? My operator is `&`.)

johnfairh:
You're right, '&' doesn't work -- it's coming in as `&amp;` to a routine that is not expecting that :/

![screen shot 2017-05-11 at 14 04 02](https://cloud.githubusercontent.com/assets/26768470/25950392/bb165c74-3652-11e7-951c-e7daa3866292.png)

reitzig:
Ahh, oh my! Editing the original post so it's immediately clear what the issue is.

Given that Swift allows Unicode identifiers, apparently some hardening has to happen.

SDGGiesbrecht:
Does the flag `--use-safe-filenames` handle the ampersand? When it is turned on, Jazzy sanitizes any characters that are bad to use in URLs (the definition of which characters are “bad” is left up to Ruby).

SDGGiesbrecht:
P.S. At least with `--use-safe-filenames` on,  I can reliably link with things like `×(_:_:)` and `√(_:)`, so this is related to HTML, not Unicode.

johnfairh:
That flag doesn't help here: the problem happens before generating the link, code is matching html-escaped text from markdown output against real declaration names. Yes general unicode is fine, it's just the html escapes & < > that cause problems AFAICT.

SDGGiesbrecht:
I just verified it. @johnfairh, you are right on all counts. I stand corrected.

[...]

Issue #789, opened 2017-04-26
Autolink in Parameter callouts


reitzig:
Autolinking does currently not work in Parameter descriptions.

johnfairh:
Fixed in master.

Issue #715, opened 2016-12-24
Autolink inside guides, readme


1ec5:
References to classes, methods, etc. should be autolinked inside the readme and guides just like in an abstract.

/cc @captainbarbosa

reitzig:
Autolinking does not work in abstract for me, either.

johnfairh:
Fixed in master.

PR #856, opened 2017-07-14
Fixes to autolinking

johnfairh:
Autolink to symbols from parameter documentation (#789) and from external markdown files, readme / documentation / custom abstract (#715).
Autolink correctly to symbols containing & < > (#805).

Fix something I broke earlier that caused undocumented parameters to appear in the parameter tables -- found while autolinking parameters.

Specs changes:
* RealmObjC - add linking from parameter descriptions
* Moya - delete empty parameters
* Alamofire - add linking from parameters and README.md
* RealmSwift - delete empty parameters; add linking from parameters
* Sierra - add linking from parameters
* MiscJazzyFeatures - add new tests

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -15,6 +15,13 @@
   [JP Simard](https://github.com/jpsim)
   [#860](https://github.com/realm/jazzy/issues/860)
 
+* Autolink from parameter documentation and from external markdown documents
+  including README.  Autolink to symbols containing & < >.  
+  [John Fairhurst](https://github.com/johnfairh)
+  [#715](https://github.com/realm/jazzy/issues/715)
+  [#789](https://github.com/realm/jazzy/issues/789)
+  [#805](https://github.com/realm/jazzy/issues/805)
+
 ## 0.8.3
 
 ##### Breaking
--- a/lib/jazzy/doc_builder.rb
+++ b/lib/jazzy/doc_builder.rb
@@ -208,6 +208,11 @@ module Jazzy
       end
     end
 
+    def self.render(doc_model, markdown)
+      html = Markdown.render(markdown)
+      SourceKitten.autolink_document(html, doc_model)
+    end
+
     # Build Mustache document from a markdown source file
     # @param [Config] options Build options
     # @param [Hash] doc_model Parsed doc. @see SourceKitten.parse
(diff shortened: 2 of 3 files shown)


jpsim:
Thanks for doing this!

Reviewing the changes to the specs, there's a somewhat-regression in Alamofire: `Notification.Name.Task` is no longer listed in the nav sidebar

Reviewing this also highlighting a pre-existing issue: how should we deal with nested types?

For example, in Alamofire, `DownloadRequest` is a class, but `DownloadRequest.DownloadOptions` is a struct, and due to the sidebar's nesting, they're both under the "Classes" heading.

image

This is especially odd with Extensions, since `Notification.Name` and `Notification.Key` are really declarations, not extensions, even though they're declared in extensions on `Notification`.

johnfairh:
On the `Notification.Name.Task` thing: this happened during #847 + #848, the extensions-of-nested-types PRs.

Before these, we were confused and the sidebar said Notification/Task which is wrong.
With the sourcekitten part, 847, the sidebar said “Notification.Name”/Task (but also “Realm.Configuration” elsewhere).
With the Jazzy part, 848, the sidebar wants to say Notification/Name/Task but can’t because it’s limited to two levels.

Maybe it would be better to avoid the unflattening for extensions of external types so it would read:
```
	Notification
		Key
	Notification.Name
		Task
```
...that way the top-level item is always a type that is being extended, even if that type is nested, and any items under it have been declared by the project.  The current version mixes up stuff in the extension (Key) with stuff already in the type that is also being extended (Name).

Well, I’ve convinced myself...

johnfairh:
Some inconclusive thoughts on nested types in general.
tl;dr: smush everything together or special-case types from extensions?

From a reader’s point of view I prefer the nested types under their parents: authors create nested types because they are strongly related to the parent type -- I'd like to read about them all together.  AF’s `DownloadRequest` is a good example there.

Agree it’s a bit odd to find a nested struct in the *Classes* category.  We could simplify these default categories into New Types/Extended Types/Functions/Global vars (or some smaller/bigger set,) merging (say) classes/structs/protocols/typealiases/enums.

As a reader, when coming in to read a project’s docs and find things from the left nav, I don’t think I initially care if a type is a struct or something else -- this is a detail of how I can use the type.  I don’t have a use case for the default layout (“today I want to learn about all `struct`s this module provides”?).  Even as a project author I’m not sure it helps me.

The merged category would be larger, for sure. Maybe keeping the category lengths down is better UI.  You probably went through this at the start of Jazzy!  I don’t know that anyone will be seriously confused by finding a nested struct in the classes category.

---

I can see how the extensions category could seem worse that the others.  I've been thinking of it as 'extended types [and their children]' category and so it has felt natural for nested types to show up there.

It’s easy enough to imagine moving docs for a new struct from an extension to the *Structs* category.  This would work nicely with Alamofire because it uses the extension solely for a namespace.  But if the project declared a type in an extension along with members for working on it/relating it to the extended type, then we'd end up with the pure extension containing just members and the new nested type elsewhere in the docs tree.  Maybe this is good - gives a bit more prominence to the new type.

johnfairh:
Thanks! Merged.

PR #875, opened 2017-08-12
Align 'jazzy' terminology with Apple usage

xwu:
When comparing documentation generated by Jazzy with those on Apple's developer site, _The Swift Programming Language_, and documentation in the Swift Project GitHub repository, some of the terminology is written out differently in public-facing contexts. Since a consistent usage can be helpful for users, this proposed change aligns Jazzy with those canonical sources.

Hopefully, this is considered a helpful modification and not duplicative of any previous proposed changes :)

> Should this be considered a productive change, I'll need guidance as to how to update the integration specs so that CI will pass.

### DIFF
--- a/lib/jazzy/source_declaration/type.rb
+++ b/lib/jazzy/source_declaration/type.rb
@@ -137,11 +137,11 @@ module Jazzy
           dash: 'Constant',
         }.freeze,
         'sourcekitten.source.lang.objc.decl.enum' => {
-          jazzy: 'Enum',
+          jazzy: 'Enumeration',
           dash: 'Enum',
         }.freeze,
         'sourcekitten.source.lang.objc.decl.enumcase' => {
-          jazzy: 'Enum Case',
+          jazzy: 'Enumeration Case',
           dash: 'Case',
         }.freeze,
         'sourcekitten.source.lang.objc.decl.initializer' => {
@@ -177,7 +177,7 @@ module Jazzy
           dash: 'Function',
         }.freeze,
         'sourcekitten.source.lang.objc.decl.struct' => {
-          jazzy: 'Struct',
+          jazzy: 'Structure',
           dash: 'Struct',
         }.freeze,
         'sourcekitten.source.lang.objc.decl.field' => {
@@ -195,27 +195,27 @@ module Jazzy
 
         # Swift
         'source.lang.swift.decl.function.accessor.address' => {
-          jazzy: 'Address Accessor',
+          jazzy: 'Addressor',
           dash: 'Function',
         }.freeze,
         'source.lang.swift.decl.function.accessor.didset' => {
-          jazzy: 'DidSet Accessor',
+          jazzy: 'didSet Observer',
           dash: 'Function',
         }.freeze,
         'source.lang.swift.decl.function.accessor.getter' => {
-          jazzy: 'Getter Accessor',
+          jazzy: 'Getter',
           dash: 'Function',
         }.freeze,
         'source.lang.swift.decl.function.accessor.mutableaddress' => {
-          jazzy: 'Mutable Address Accessor',
+          jazzy: 'Mutable Addressor',
           dash: 'Function',
         }.freeze,
         'source.lang.swift.decl.function.accessor.setter' => {
-          jazzy: 'Setter Accessor',
+          jazzy: 'Setter',
           dash: 'Function',
         }.freeze,
         'source.lang.swift.decl.function.accessor.willset' => {
-          jazzy: 'WillSet Accessor',
+          jazzy: 'willSet Observer',
           dash: 'Function',
         }.freeze,
         'source.lang.swift.decl.function.operator' => {
@@ -247,11 +247,11 @@ module Jazzy
           dash: 'Class',
         }.freeze,
         'source.lang.swift.decl.function.constructor' => {
-          jazzy: 'Constructor',
+          jazzy: 'Initializer',
           dash: 'Constructor',
         }.freeze,
         'source.lang.swift.decl.function.destructor' => {
-          jazzy: 'Destructor',
+          jazzy: 'Deinitializer',
           dash: 'Method',
         }.freeze,
         'source.lang.swift.decl.var.global' => {
@@ -259,15 +259,15 @@ module Jazzy
           dash: 'Global',
         }.freeze,
         'source.lang.swift.decl.enumcase' => {
-          jazzy: 'Enum Case',
+          jazzy: 'Enumeration Case',
           dash: 'Case',
         }.freeze,
         'source.lang.swift.decl.enumelement' => {
-          jazzy: 'Enum Element',
+          jazzy: 'Enumeration Element',
           dash: 'Element',
         }.freeze,
         'source.lang.swift.decl.enum' => {
-          jazzy: 'Enum',
+          jazzy: 'Enumeration',
           dash: 'Enum',
         }.freeze,
         'source.lang.swift.decl.extension' => {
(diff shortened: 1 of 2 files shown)


SDGGiesbrecht:
@xwu, I like your use of real English words better than the existing code abbreviations, but you will have to wait for the maintainers’ thoughts.

The risk involved is that the way you have written the changes, they also affect the URLs. I do not see anywhere that Jazzy has made a documented guarantee that URLs would remain stable, but I know many people use them as if that were the case.

The optimal solution would be to output a file at each old URL which would redirect to the corresponding new URL. The question is whether it is worth the effort. Wait and see what the folks at Realm have to say.

johnfairh:
I agree we should use consistent correct terminology, it helps everybody.

The URLs that change here are for enums, structs, and type aliases.  I don't know how big an issue this is; users could create links if really a problem.  @SDGGiesbrecht can you say a bit more about the broken use cases in practice that you're familiar with?  It would be easy enough to have a different version of `plural_name` etc. for building the URL that preserved the legacy naming but we would be stuck with that code forever.

(I was going to complain about the space in 'Type Alias' but turns out we already have things like 'Global Variables' and 'Other Classes' which work well enough.)

@xwu did you intend to change the Objective-C spelling of 'enum' and 'structure' as well as Swift?  It does probably makes sense to be consistent across the languages.

I created [this branch](https://github.com/realm/jazzy-integration-specs/tree/jf-jazzy-pr-875) to show the specs with the changes.

SDGGiesbrecht:
@johnfairh,

I myself do not care if URLs change. It was just something I remember being asked to consider back when we were working on removing dangerous characters from URLs. One of the reasons people wanted the change to be opt‐in was because they had hard‐coded links from read‐me files and the like, which would have broken without them noticing:

```markdown
See [MyStructure](mydomain.tld/docs/Structs/MyStructure.html).
```

And who knows who else has linked to your documentation without you knowing?

I repeat though, that I am of the opinion that a link is the linkers responsibility to maintain, not the linkee.

There are three ways of going about it:

1) Let the URLs change. (What this pull request does, and my personal preference.)
2) Decouple the URLs from the link text. (i.e. the link says “Structures” but they still go in a folder called `Structs`).
3) Some automated redirection scheme to send the old URLs to the new ones.

xwu:
@johnfairh I do believe my proposed changes include parallel ones for Obj-C. Apologies for not calling it out explicitly.

johnfairh:
Ok good, just checking.  I think we should preserve the directory names at least; change is not that bad and may save some user frustration - [patch](https://gist.githubusercontent.com/johnfairh/ea13a98258e2e4da103d5a3322f60ee2/raw/8a1d89936fdfd7f97f2e84c51ca8d92931ad8e49/keep_type_dir_names.patch).

xwu:
@johnfairh Shall I roll that into this PR, or would you like to open another one for preservation of directory names?

johnfairh:
I think it'll be clearer if I do the directory name change as a follow-up.  Could you add a [change-log entry](https://github.com/realm/jazzy/blob/master/CONTRIBUTING.md) to this one though please?

lib/jazzy/source_declaration/type.rb
```diff
@@ -195,27 +195,27 @@ def ==(other)
 
         # Swift
         'source.lang.swift.decl.function.accessor.address' => {
-          jazzy: 'Address Accessor',
+          jazzy: 'Addressor',
           dash: 'Function',
         }.freeze,
         'source.lang.swift.decl.function.accessor.didset' => {
-          jazzy: 'DidSet Accessor',
+          jazzy: 'didSet Observer',
```
pcantrell: Headings should be capitalized even if the corresponding language construct is not. (Consider that the “Class” heading above is capitalized even though lowercase `class` is a language keyword.)

If “DidSet” is bothersome, then this heading should not use a language keyword at all — following the precedent of Enumeration, Structure, and Type Alias.

xwu: As the description states, all of these changes have been made to align with Apple's developer site, _The Swift Programming Language_, and documentation in the Swift GitHub repository.

In _TSPL_, the usage is always "`didSet` observer" with the keyword in code voice and without changes in capitalization. In these source documents, `didSet` and `willSet` observers are _always_ referred to by their language keyword, while no other entities on our list are _ever_ referred to by their language keyword. I have adhered strictly to precedent and exercised no editorial judgment in this.

pcantrell: Before posted that comment, I looked at _The Swift Programming Language_, and was unable to find any examples of `didSet` or `willSet` used in titles or section headings. Can you find an example where they do?

> I have adhered strictly to precedent and exercised no editorial judgment in this.

The rationale in your reply — and this entire thread — _is_ editorial judgement. 😉

xwu: There are no uses of the term in titles or section headings (of which I'm aware), but you will find that the entity is referred to as the `didSet` observer throughout _TSPL_, and that `didSet` is never spelled in any other fashion even at the beginning of sentences.

I, for one, am entirely comfortable with the notion that the term, in its capitalized form, remains "didSet," like "iPhone" or "pH." Certainly, it is never spelled "DidSet" in Apple or swift.org documentation.

SDGGiesbrecht: Nice catch, @pcantrell, @xwu 

To my eyes, either of the first two rows looks good, but the third looks ugly and unnatural:

| | ...normal case... | Sentence case. | Title/Header Case |
| - | - | - | - |
| ✓ Proper English | ...did‐set observer... | Did‐set observer. | Did‐Set Observer |
| ✓ Rich Text, Code Voice | ...`didSet` observer... | `didSet` observer. | `didSet` Observer |
| ✗ Raw Text, Code Voice | ...didSet observer... | didSet observer. | didSet Observer |

(“DidSet” would be even worse...)

Since the second is what *The Swift Programming Language* does, is it worth trying `<code>didSet</code> Observer` in that field? Does that cause other things to break? How does it look like with Jazzy’s CSS?

The same goes for “will‐set observer”.

Also, should “ivar” be changed to “instance variable” for consistency, since we are speaking English not code?

johnfairh: In practice I think this is moot for jazzy right now: users cannot not see these particular strings.

Property observers are not reported by SourceKit as part of document structure so jazzy does not see them and cannot use their type names as markdown or in CLI warning messages.  (Related, they are undocumentable today: try getting Xcode quick help to recognise a doc comment.)

If Swift were to change to include accessors in this view then we would need to make a few changes to the pipeline; if we decided to show them separately in the docs then those would include making the UI and its spelling acceptable to everyone.

The 'Ivar' represents a libclang `CXCursor_ObjCIvarDecl`.  This doesn't sound like something that can occur at the top level of a header file, so probably not a string a user can see in jazzy.  I guess it would be in keeping with the other changes here to change it though.

xwu: Yes, the point is fairly moot.

However, when it does become possible for this term to be user-facing, the _TSPL_-compliant way to refer to a `didSet` observer in a heading is unambiguously "didSet Observer" without rich text. This is evidenced by existing headings in _TSPL_ such as "The self Property" (note the lack of capitalization and lack of monospaced formatting for the word "self").

I considered and left "ivar" alone for this particular PR, as I could find no evidence of its spelling in Apple documentation. However, "instance variable" seems reasonable.

johnfairh: @pcantrell, @SDGGiesbrecht - are you OK with this PR as-is, leaving the strings consistent with Swift and not user-visible?  Or still prefer to leave didSet/willSet unchanged?

(['instance variable' mentions](https://developer.apple.com/library/content/documentation/Cocoa/Conceptual/ProgrammingWithObjectiveC/EncapsulatingData/EncapsulatingData.html))

xwu: @johnfairh Instance variable it is then :)

SDGGiesbrecht: For declaration names that are not user‐visible, everything I have said is irrelevant.

You can even call it `ḓïđŠéț` if you want. ☺

johnfairh:
Merged via #884 - thanks for the contribution.

PR #884, opened 2017-09-26
Align terminology with Apple usage

johnfairh:
PR to merge #875 with working specs and patch to preserve directory names in URLs.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,7 @@
 
 * Align jazzy terminology with Apple usage.  
   [Xiaodi Wu](https://github.com/xwu)
+  [John Fairhurst](https://github.com/johnfairh)
 
 ##### Bug Fixes
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,7 +6,8 @@
 
 ##### Enhancements
 
-* None.
+* Align jazzy terminology with Apple usage.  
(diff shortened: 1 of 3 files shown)


jpsim:
Thanks for driving this! Really appreciate all your hard work to move this project forward lately.

FYI, some URLs appear to have changed though. Here's the integration specs diff from 0.8.3 to master: https://github.com/realm/jazzy-integration-specs/compare/9867fcd8bb20d842c8666fe39330bf95b394ce63...6654c65db15555e9b8b93683ca3e16235a391665

johnfairh:
Right; the idea was to preserve the directory names, so root/Structs/MyStruct.html is still there, but not the summary pages (so root/Structs.html has become root/Structures.html).  Because less code to maintain (for sure) and users much less likely to have links into the summary pages (probably??).  Do you think it's worth preserving the summary page names too?

jpsim:
If we keep as-is, the changelog entry should be moved to the "Breaking Changes" section, with an explanation of what's breaking (i.e. what you just wrote)

Personally, it seems inconsistent to change it if it's in one location in the URL, but not another. My slight personal preference would be to not break any URLs whatsoever from the 0.8.3 release.

johnfairh:
Ok, thanks.  Yes one of those times where writing out the design again makes me less convinced by it.  TBC then.

PR #886, opened 2017-09-27
Rename category files to restore URL naming

johnfairh:
This PR restores all URLs to be the same as before #884 while keeping the new terminology in text.
[Spec diffs from before any terminology changes to this PR.](https://github.com/realm/jazzy-integration-specs/compare/4c740e7a5cefd59cb775f8fbe995b99d7c9d7f38...jf-fix-all-urls)

I realized the current code in master has broken custom abstract markdown for files like Struct[ure]s.md, one more reason to do this.

I had to change the moustache template though which hard-coded filename == nav text.  Does this mean this is a still a breaking change because users with custom themes will need to refresh that file?

### DIFF
--- a/lib/jazzy/doc_builder.rb
+++ b/lib/jazzy/doc_builder.rb
@@ -46,6 +46,7 @@ module Jazzy
         end
         {
           section: doc.name,
+          url: doc.url,
           children: children,
         }
       end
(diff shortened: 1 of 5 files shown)


jpsim:
Merged in 817816c & 8a5f9ac. Thanks!

johnfairh:
Ty for merge.

Issue #768, opened 2017-04-15
Lazy properties documentation includes unneeded code


ilyapuchka:
When documenting lazy properties jazzy can include it's actual code as part of it's declaration documentation.

Example - https://github.com/krzysztofzablocki/Sourcery/blob/master/docs/Classes/TypesReflectionBox.html#L176 In this example `=` is included as part of declaration documentation. 

If the property body is defined in the same line it will be completely included in the declaration documentation. 

```swift 
public lazy var classes: [Class] = { return self.types.flatMap { $0 as? Class } }()
``` 

This code produces the following html output:

```html
<pre class="highlight"><code><span class="kd">public</span> <span class="kd">lazy</span> <span class="k">var</span> <span class="nv">classes</span><span class="p">:</span> <span class="p">[</span><span class="kt"><a href="../Classes/Class.html">Class</a></span><span class="p">]</span> <span class="o">=</span> <span class="p">{</span> <span class="k">return</span> <span class="k">self</span><span class="o">.</span><span class="n">types</span><span class="o">.</span><span class="n">flatMap</span> <span class="p">{</span> <span class="nv">$0</span> <span class="k">as?</span> <span class="kt"><a href="../Classes/Class.html">Class</a></span> <span class="p">}</span> <span class="p">}()</span></code></pre>
```

reitzig:
Related: #776, #777

johnfairh:
Fixed via #902.

PR #902, opened 2017-11-07
Improve Swift declarations using SourceKit's version

johnfairh:
_Still working through specs and tweaking things, appreciate feedback about any of this especially around how `@attribute`s are handled._

This changes Swift declarations to use the compiler’s round-tripped version instead of the raw parsed source code in many cases.  Improvements:
* Show { get } / { get set } for properties in protocols, for subscripts, and for computed properties
* If code says `var fs = “Fred”` then show decl `var fs: String`
* Don’t show unwanted code fragments that confused the sourcekitten parser (eg. #768)
* Show all declaration attributes (eg. `@discardable, @objc, @available`) 
* Standardize formatting (eg. whitespace)

Still prefer the parsed declaration when it is better:
* Functions with default parameters.  Values not available from compiler, SR-2608, perhaps fixed in Swift 5.
* Declarations arranged over multiple lines - presume the user has formatted these nicely so keep them.  Should fix to be adaptive in Jazzy html/js some day.
* A ?bug to do with `@escaping / @autoclosure` - SR 6321, try + fix for Swift 4.1.

Re. decl attributes, I’ve chosen to stack these each on their own line, to avoid very long lines. I think they’re all useful in docs --- API users may legitimately want to know these?  Eg:

screen shot 2017-10-27 at 14 45 55

screen shot 2017-10-27 at 15 02 36

screen shot 2017-10-27 at 18 05 11
[this last fabricated of course but you get the idea...]

Jazzy should handle `@available` in a more visual way but for now I think it’s an improvement to pass through what the author wrote?  The highlighting is a bit ugly; there’s a rouge PR to improve it.

Notes on spec changes in the specs PR.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,12 @@
   [Nick Fox](https://github.com/nicholasffox)
   [#949](https://github.com/realm/jazzy/issues/949)
 
+* Improve Swift declarations to look more like the Xcode Quick Help version,
+  for example including { get set }, and include all attributes.  
+  [John Fairhurst](https://github.com/johnfairh)
+  [#768](https://github.com/realm/jazzy/issues/768)
+  [#591](https://github.com/realm/jazzy/issues/591)
+
 ##### Bug Fixes
 
 * Preserve `MARK` comment headings associated with extensions and enum cases.  
--- a/lib/jazzy/source_declaration.rb
+++ b/lib/jazzy/source_declaration.rb
@@ -49,6 +49,14 @@ module Jazzy
       namespace_path.map(&:name).join('.')
     end
 
+    # :name doesn't include any generic type params.
+    # This regexp matches any generic type params in parent names.
+    def fully_qualified_name_regexp
+      Regexp.new(namespace_path.map(&:name)
+                               .map { |n| Regexp.escape(n) }
+                               .join('(?:<.*?>)?\.'))
+    end
+
     # If this declaration is an objc category, returns an array with the name
     # of the extended objc class and the category name itself, i.e.
     # ["NSString", "MyMethods"], nil otherwise.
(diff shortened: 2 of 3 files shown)


johnfairh:
Going over the specs this all looks fine with various fixes and improvements.  One part of 'standardise formatting' to call out: generic parameter constraints are always displayed in the `where` clause.  So source code:
screen shot 2017-11-10 at 10 56 39
becomes:
screen shot 2017-11-10 at 10 55 42

jpsim:
So we _used_ to do this a long time ago, but lost too much of the formatting from the original source to be worthwhile.

Also, I _think_ there are printing options for SourceKit (at least internally to the Swift compiler) that allow printing out default parameter _values_ rather than just `default`: https://github.com/apple/swift/blob/39c3fae1df18f2804b03b37c4446b57ca20effb2/include/swift/AST/PrintOptions.h#L145-L146

johnfairh:
Yes I kept finding small problems needing workarounds; thought I'd keep going though.

That print option just controls whether you get `= default` or nothing at all :/  To make it work well I think we need Swift 5 [+] to move the default expression into the module interface, currently they're not in the serialised AST.

jpsim:
@johnfairh what would you like to do with this? If it's not a 100% improvement on existing behavior, but improves some aspects, I wouldn't be opposed to having this behavior as a configuration flag.

Let me know if you want me to review again.

johnfairh:
@jpsim I *think* the changes this PR makes are all improvements.  If it spots a declaration where the compiler's version is worse (eg. default params,) then it uses the parsed version as today.

Let me rebase + retest...

johnfairh:
Fiiiiinally back to this.  Reworked code a bit to fix the two problems from before:
* All attributes are now picked up even if parsed version of declaration is used;
* Worked around Rouge formatting bug with @available + quoted strings + parens

Worth a review @jpsim when you have a chance -- I think it improves everything.

johnfairh:
Thanks - merging.  Code here definitely benefited from the gestation period.

PR #914, opened 2017-11-29
Making the "Undocumented" text for functions that are not documented. configurable

akhillies:
Small change: made the text for undocumented text configurable

1) Added a config option: undocumented_text
2) used that config option as the description text if set for undocumented functions

This should not change the default behavior (if you do not set the option) and I verified it worked with the option in both command line and config file.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,7 +6,10 @@
 
 ##### Enhancements
 
-* None.
+* Added a config option (--undocumented-text) to set the default text for
+  undocumented code.
+  [Akhil Batra](https://github.com/akhillies)
+  [#913](https://github.com/realm/jazzy/issues/913)
 
 ##### Bug Fixes
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,7 @@
   [Akhil Batra](https://github.com/akhillies)
   [#913](https://github.com/realm/jazzy/issues/913)
 
+
 ##### Bug Fixes
 
 * None.
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,8 +6,8 @@
 
 ##### Enhancements
 
-* Added a config option (`--undocumented-text UNDOCUMENTED_TEXT`) to set the default text for
-  undocumented symbols.
+* Added a config option (`--undocumented-text UNDOCUMENTED_TEXT`) to set the
+  default text for undocumented symbols.  
   [Akhil Batra](https://github.com/akhillies)
   [#913](https://github.com/realm/jazzy/issues/913)
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,8 +6,8 @@
 
 ##### Enhancements
 
-* Added a config option (--undocumented-text) to set the default text for
-  undocumented code.
+* Added a config option (`--undocumented-text UNDOCUMENTED_TEXT`) to set the default text for
+  undocumented symbols.
   [Akhil Batra](https://github.com/akhillies)
   [#913](https://github.com/realm/jazzy/issues/913)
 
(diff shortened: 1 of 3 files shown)


akhillies:
If I still need to run the integration tests, can I please have access to push and run them?

akhillies:
what does "Perceived complexity for process_undocumented_token is too high" mean and how do I fix that?

CHANGELOG.md
```diff
@@ -6,7 +6,10 @@
 
 ##### Enhancements
 
-* None.
+* Added a config option (--undocumented-text) to set the default text for
+  undocumented code.
```
johnfairh: Per CONTRIBUTING.md, need two spaces after the end of the description please :)

lib/jazzy/config.rb
```diff
@@ -328,7 +328,13 @@ def expand_path(path)
         raise '--assets-directory is deprecated: use --theme instead.'
       end
 
-    # rubocop:enable Style/AlignParameters
+    config_attr :undocumented_text,
+      command_line: '--undocumented-text UNDOCUMENTED_TEXT',
+      description: 'Default text for undocumented functions and  properties. '\
```
johnfairh: "for undocumented symbols" ?

lib/jazzy/sourcekitten.rb
```diff
@@ -48,7 +48,8 @@ def autolink_block(doc_url, middle_regex, after_highlight)
 module Jazzy
   # This module interacts with the sourcekitten command-line executable
   module SourceKitten
-    @undocumented_abstract = Markdown.render('Undocumented').freeze
+    @undocumented_abstract = Markdown.render(Config.instance.undocumented_text)
```
johnfairh: bcos you set the default to 'Undocumented' we can always use the config setting without any "if"s, this part can be like
```ruby
def self.undocumented_abstract
  @undocumented_abstract ||= Markdown.render(Config...)
end
```

lib/jazzy/sourcekitten.rb
```diff
@@ -282,7 +283,11 @@ def self.process_undocumented_token(doc, declaration)
       if objc || should_mark_undocumented(doc['key.kind'], filepath)
         @stats.add_undocumented(declaration)
         return nil if @skip_undocumented
-        declaration.abstract = @undocumented_abstract
```
johnfairh: ...and this can be `= undocumented_abstract` without ifs to upset rubocop.

johnfairh:
Looks good, thanks for doing it -- you don't need any spec changes here because the default behaviour is unchanged.  The 'complexity' message is from rubocop, a linter, I suggested some review changes to make it happier -- you can run it locally before commit by doing `rake rubocop`.

lib/jazzy/config.rb
akhillies: good call

CHANGELOG.md
akhillies: whoops, sorry about that

lib/jazzy/sourcekitten.rb
akhillies: thank you :)

akhillies:
great fixed! (also thanks for the rubocop tip, I didn't know how to run it without the CI system)

jpsim:
Thanks for the PR!

> I didn't know how to run it without the CI system

```
$ # Install dependencies
$ bundle install
$ # Run Ruby style linter
$ bundle exec rake rubocop
$ # Run all tests
$ bundle exec rake spec
```

CHANGELOG.md
```diff
@@ -6,7 +6,11 @@
 
 ##### Enhancements
 
-* None.
+* Added a config option (--undocumented-text) to set the default text for
```
jpsim: backticks around `--undocumented-text`

jpsim:
Lovely.

PR #917, opened 2017-12-01
Adds a modern Apple docs theme

HarshilShah:
This is a modification of the existing `apple` theme within Jazzy, to make it look similar to the current Apple docs style.

There are some deviations from the exact Apple style, to accommodate for the difference in documentation styles between Apple and (at least in my experience) the average OSS project, but the gist remains the same. It's a single column, responsive layout which resembles Apple's own docs as of this writing.

I didn't know how big the changes would be, so rather than submit a massive PR with a lot of commits, I instead developed it in a [separate repo](https://github.com/HarshilShah/Jony), and am PRing only the final version here.


hollanderbart:
Cool!

johnfairh:
Nice!  I'll stick in a screenshot below for people to look at, don't know if you have a site generated somewhere?

Have you seen the searchbar function in the `fullwidth` theme?  Would be nice to include that (maybe as follow-on), is a big functional advantage of that theme today.

Not a strong opinion from not a visual designer: I wonder if the background colour for the twisty expansion would be better as white (like fullwidth) instead of grey?

Is 'jony' what this is actually called or a hat-tip?  'apple2017' or something would be more boring but might help people find it.

There are a couple code edits needed starting [here](https://github.com/realm/jazzy/blob/893a215b050dbd8a8a5495d1007391cc7048ac7b/lib/jazzy/config.rb#L295) to support a new theme.  Plus changelog + readme.

screen shot 2017-12-06 at 12 32 53

HarshilShah:
@johnfairh Sorry about the demo site; it's just that none of my personal projects are complex enough (seriously there are 3 files in the biggest one). I've been developing by testing against IGListKit, let me see if I can get a fork of that to render to GitHub Pages.

Haven't seen the search functionality in the `fullwidth` theme, will take a look.

Grey was just in keeping with Apple's own design.

Jony just seemed like a natural second choice since Apple was taken. Apple2017 feels like its asking to be outdated in <1 month now though. This seems like the new permanent direction for Apple docs, not a trend, so it made sense to name it after the guy responsible, but I'm a-okay with another name.

HarshilShah:
So here are the regenerated [IGListKit docs](https://harshilshah.github.io/IGListKit/)

amorde:
@HarshilShah Not sure if this is my specific setup, but the IGListKit doc demo site has some clipping in the nav bar with a scroll indicator on the right side

iglistkit_new_theme

This was using Safari `11.0.2 (12604.4.7.1.4)` on macOS `10.12.6`

jpsim:
Merged in #955 with the little few things that were left to do. Thanks @HarshilShah and everyone who reviewed.

Issue #878, opened 2017-09-04
Overloaded generic operator functions use the same filename


johnfairh:
Operator functions with a doc comment and at least one generic type parameter end up with a spurious child documentation entry for the _first_ generic type parameter.

Jazzy should suppress these -- neither Xcode nor swift know how to document them yet so nothing is lost.  The impact is shown in #877: if you have multiple such operator functions with the same name but different params then the same filename is used for all their html pages, which is bad.  Without the generic type param, the function won't get its own html page, so no problem.

SDGGiesbrecht:
@johnfairh, I remember doing some related work back in December (2016). At that point Jazzy even threw `undocumented` errors for those generic type parameters, so I submitted a fix that filtered them out. The key line is [here](https://github.com/realm/jazzy/blob/a02fe0a86e02e5e0d67e96a05d5040478a8a36a3/lib/jazzy/sourcekitten.rb#L270). At the time we did not realize there were problems with the documentation output too. Maybe all that is needed is to apply the same filter while constructing the documentation tree?

johnfairh:
Yep it's an easy fix.

johnfairh:
fixed in master.

PR #922, opened 2017-12-06
Don't create documentation for generic type params

johnfairh:
From #878.  Generic type param nodes appear intermittently in the doc structure as children of functions and cause extra html pages to be created.  Just skip them.

Spec changes show effect on misc_jazzy_features - with an extra change there to continue exercising the safe-filenames feature.


johnfairh:
Rubocop appeased + merged.

PR #923, opened 2017-12-07
Added a config option to generate only Swift or Objective-C declarations when generating Objective-C docs.

ulukaya:

### DIFF
--- a/lib/jazzy/config.rb
+++ b/lib/jazzy/config.rb
@@ -129,8 +129,8 @@ module Jazzy
 
     config_attr :hide_declarations,
       command_line: '--hide-declarations [swift|objc] ',
-      description: 'Hide language declarations when generating Objective-C '\
-                   'framework.',
+      description: 'Hide language declarations when generating docs for an '\
+                   'Objective-C framework.',
       default: ''
 
     config_attr :config_file,
--- a/lib/jazzy/config.rb
+++ b/lib/jazzy/config.rb
@@ -129,7 +129,8 @@ module Jazzy
 
     config_attr :hide_declarations,
       command_line: '--hide-declarations [swift|objc] ',
-      description: 'Hide language declarations when generating Objective-C framework.',
+      description: 'Hide language declarations when generating Objective-C '\
+                   'framework.',
       default: ''
 
     config_attr :config_file,
--- a/lib/jazzy/config.rb
+++ b/lib/jazzy/config.rb
@@ -127,6 +127,11 @@ module Jazzy
       description: 'The SDK for which your code should be built.',
       default: 'macosx'
 
+    config_attr :hide_declarations,
+      command_line: '--hide-declarations [swift|objc] ',
+      description: 'Hide language declarations when generating Objective-C framework.',
+      default: ''
+
     config_attr :config_file,
       command_line: '--config PATH',
       description: ['Configuration file (.yaml or .json)',
--- a/lib/jazzy/doc_builder.rb
+++ b/lib/jazzy/doc_builder.rb
@@ -340,13 +340,12 @@ module Jazzy
                     else
                       item.declaration
                     end
-      other_language_declaration = item.other_language_declaration if
-          Config.instance.hide_declarations == ''
       {
         name:                       item.name,
         abstract:                   abstract,
         declaration:                declaration,
-        other_language_declaration: other_language_declaration,
+        other_language_declaration: (item.other_language_declaration if
+            Config.instance.hide_declarations == ''),
         usr:                        item.usr,
         dash_type:                  item.type.dash_type,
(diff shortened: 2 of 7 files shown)


ulukaya:
@jpsim , what do you think?

lib/jazzy/config.rb
```diff
@@ -127,6 +127,12 @@ def expand_path(path)
       description: 'The SDK for which your code should be built.',
       default: 'macosx'
 
+    config_attr :hide_declarations,
+      command_line: '--hide-declarations [swift|objc] ',
+      description: 'Hide language declarations when generating Objective-C '\
+                   'framework.',
+      default: ''
+
```
johnfairh: Be good to police the user enters one of these strings + nothing else, and also that `--objc` is set -- I know jazzy doesn't do much of that today but that does cause user confusion.  (also wording, "...generating *docs for an* Objective...")

lib/jazzy/doc_builder.rb
```diff
@@ -329,17 +329,28 @@ def self.gh_token_url(item, source_module)
...
+    # rubocop:disable Metrics/PerceivedComplexity
+    # rubocop:disable Metrics/MethodLength
+
     # Build mustache item for a top-level doc
     # @param [Hash] item Parsed doc child item
     # @param [Config] options Build options
     def self.render_item(item, source_module)
       # Combine abstract and discussion into abstract
       abstract = (item.abstract || '') + (item.discussion || '')
+      declaration = if Config.instance.hide_declarations == 'objc'
```
johnfairh: Probably neater to pull these out into eg. `display_declaration` methods on `SourceDeclaration`.

Expected to see a similar change to `DocBuilder.document()` - doesn't the current code always show the objc declaration at the top of each page?

lib/jazzy/sourcekitten.rb
```diff
@@ -361,7 +361,12 @@ def self.make_source_declarations(docs, parent = nil)
         declaration.parent_in_code = parent
         declaration.type = SourceDeclaration::Type.new(doc['key.kind'])
         declaration.typename = doc['key.typename']
-        current_mark = SourceMark.new(doc['key.name']) if declaration.type.mark?
+        documented_name = if Config.instance.hide_declarations == 'objc'
+                            doc['key.swift_name']
+                          else
+                            doc['key.name']
+                          end
```
johnfairh: Sometimes there is no Swift decl for an ObjC decl, ie. cursorInfo returns nothing.  Not sure how well code will react to having nil for `documented_name` hereon down.

Would be nice to factor the config queries into config to cut down on the number of literal strings scattered around!

johnfairh:
I added a few comments - direction looks right to me - would be worth running the realm obj-c project (under specs directory) through in both new modes + seeing what the docs look like, there is a good deal of obj-c coverage there.

lib/jazzy/config.rb
ulukaya: updated the description. Those are all good to have, but I don't see any check at all on config today. Maybe I can do a separated PR to do it once for all. I also only check the existence of 'swift or objc' in many places and fix others too, so putting a any other string will have no effect. It'll simply ignore.

lib/jazzy/doc_builder.rb
ulukaya: done.

that behavior is actually on mustache templates. It doesn't show other declaration unless there is declaration. It picks up first declaration name from "language" and 2nd declaration is Swift by default. That's why I change the language at doc.rb to simplify the change.

lib/jazzy/sourcekitten.rb
ulukaya: Done.

ulukaya:
@johnfairh thanks for the review. should be better now.

ulukaya:
screen shot 2017-12-10 at 9 33 17 am
screen shot 2017-12-10 at 9 37 01 am
screen shot 2017-12-10 at 9 39 49 am

johnfairh:
I'm going to bang on about `DocBuilder.document()` again.

As I understand the goal of this PR, the idea of `--hide-declarations objc` is to hide the Objective-C declarations and show Swift ones, where they exist.  So I still think this routine needs a change to do  that -- by eg. using your new `display_declaration`.  Otherwise we get an index page like this:

screen shot 2017-12-11 at 18 31 36
...where *See more* links to a page full of Swift except the very top that says:
screen shot 2017-12-11 at 18 31 55
...which is wrong, right?


Only other thing: be good to add a couple of lines mentioning the feature to the 'Objective-C' part of the readme.

ulukaya:
@johnfairh fixed that and the enums. Thanks.

ulukaya:
@jpsim good to merge?

jpsim:
Made a few small changes and merged as b139d92. Thanks.

PR #929, opened 2017-12-19
Migrate to Circle CI 2.0 and parallelize jobs

jpsim:


jpsim:
Runs in ~8 minutes total now instead of ~13 minutes before.

PR #930, opened 2017-12-20
Update to latest Rouge for better highlighting

johnfairh:
This PR supports latest Rouge covering Swift 3 (including `open`, `fileprivate`) and all kinds of other things.

The jazzy code changes are to handle changes to the Rouge API/behaviour:
1. None of the built-in formatters generate exactly the HTML + CSS that jazzy expects.
2. Rouge 3 stopped being able to auto-guess Objective-C.  To work around this the PR explicitly sets the default language for markdown code blocks.

The spec diffs are voluminous due to all the highlighting changes, plus all the code-highlight blocks are now styled identically: previously the declaration highlights did not include the language name as a CSS style, but the markdown blocks did; now they all do.

Can see examples of the highlighter in eg. the index pages; [RLMArray addNotificationBlock] docs show the auto-language part.

I tested this with rouge 2.0.6 as well as 3.0.0 (latest) and the only spec differences were to highlighting of `shell` code.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -16,6 +16,11 @@
   [Ibrahim Ulukaya](https://github.com/ulukaya)
   [#828](https://github.com/realm/jazzy/issues/828)
 
+* Automatically use Swift or Objective-C syntax highlighting for code blocks
+  in documentation comments.  Improve Swift highlighting with latest Rouge.  
+  [John Fairhurst](https://github.com/johnfairh)
+  [#218](https://github.com/realm/jazzy/issues/218)
+
 ##### Bug Fixes
 
 * Fix Swift declarations when generating Objective-C docs for generic types.  
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -6,7 +6,7 @@ PATH
       mustache (~> 0.99)
       open4
       redcarpet (~> 3.2)
-      rouge (~> 1.5)
+      rouge (>= 2.0.6, < 4.0)
       sass (~> 3.4)
       sqlite3 (~> 1.3)
       xcinvoke (~> 0.3.0)
@@ -130,7 +130,7 @@ GEM
     rb-inotify (0.9.10)
       ffi (>= 0.5.0, < 2)
     redcarpet (3.4.0)
-    rouge (1.11.1)
+    rouge (3.0.0)
     rubocop (0.49.0)
       parallel (~> 1.10)
       parser (>= 2.3.3.1, < 3.0)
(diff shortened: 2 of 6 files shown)


SDGGiesbrecht:
@johnfairh, I’m curious—how difficult would it be to filter out the `<span class="err">` instances Rouge produces, especially the ones that split up grapheme clusters?

For example, run...

```shell
git clone https://github.com/SDGGiesbrecht/SDGLogic
cd SDGLogic
swift package generate-xcodeproj
jazzy
open docs/index.html
```

...then scroll down to the “Example Usage” section and look what happened to `≠`—it got split in two.

(This problem is not new and I know it is really Rouge’s fault, but I though I would ask anyway while we are dealing with the related code. P.S. I already looked into fixing Rouge, but it parses UTF‐8 units, not Unicode scalars, so it is more or less hopeless.)

johnfairh:
Thanks - merging...

@SDGGiesbrecht I don't know enough about unicode to figure out what the right fix is: why rouge doesn't think \neg etc. aren't valid identifiers -- do you think it is just not fixable or fixable with a re-do of how the Swift lexer [recognises identifiers in Ruby strings](https://github.com/jneen/rouge/blob/9c5f98944b4ecc84eba892463d687d0042b45d22/lib/rouge/lexers/swift.rb#L14)?

I did think while fixing backticks/keypath that the "err" visual style is not that useful here, be better to do something quieter.  In that spirit stripping out the 'err' spans and merging their content with the previous one is not perfect but works acceptably I reckon:

screen shot 2017-12-21 at 11 22 41

better than

screen shot 2017-12-21 at 11 38 37

Quick patch [here](https://gist.github.com/johnfairh/1ab0e80d08cfe292433217cc9fb3e92a) if you care to tidy it up - need the same thing/encapsulate in `Highlighter.highlight` too + handle no preceding span.  Or can look at it next year.

SDGGiesbrecht:
> In that spirit stripping out the 'err' spans and merging their content with the previous one is not perfect but works acceptably...

That is exactly what [I currently do](https://github.com/SDGGiesbrecht/Workspace/blob/eb86890529f5e469e3159577f9f69cb89345cd5c/Sources/WorkspaceLibrary/External%20Tools/Third%20Party/Jazzy.swift#L177-L191) to post‐process Jazzy’s output.

***

> Do you think it is just not fixable or fixable with a re‐do of how the Swift lexer recognises identifiers in Ruby strings?

A long time ago when I first looked, it seemed it would require a complete rewrite of Rouge’s treatment of strings. Now it looks more doable. I do not know whether I was originally mistaken, or if Rouge has evolved significantly since then.

***

Since I post‐process anyway, I am in no hurry to get it fixed. I just thought it would be helpful if someone more Ruby‐minded had it in the back of their head as well. Thanks for all your work, @johnfairh.

SDGGiesbrecht:
I just took a crack at Rouge again, replacing the oversimplified regexes with full definitions from the [official grammar](https://developer.apple.com/library/content/documentation/Swift/Conceptual/Swift_Programming_Language/zzSummaryOfTheGrammar.html), but after several hours of chasing infinite loops, I gave up again.

jpsim:
Thanks for pursuing it nonetheless @SDGGiesbrecht

PR #933, opened 2018-01-09
Update specs to build with Rouge 3.1

johnfairh:
Bundler has picked up rouge 3.1 which has changed a couple of things.


jpsim:
Thanks!

Issue #941, opened 2018-02-07
Jazzy fails to install via ruby gem.


JacobKingDev:
The ruby gem installation fails. I have taken debugging steps and re-installed dependencies etc but have been unable to resolve the situation. The issue seems to be related to `libffi` being built for the wrong CPU architecture?

I'm running OSX High Sierra 10.13.1, Xcode 9.2, Swift 4.0, zsh

Please see below for the error log:

```
libtool: link: rm -fr  .libs/libffi_convenience.a .libs/libffi_convenience.la
libtool: link: ar cru .libs/libffi_convenience.a src/.libs/prep_cif.o src/.libs/types.o src/.libs/raw_api.o src/.libs/java_raw_api.o src/.libs/closures.o src/x86/.libs/ffi64.o src/x86/.libs/unix64.o src/x86/.libs/ffiw64.o src/x86/.libs/win64.o
/Applications/Xcode.app/Contents/Developer/Toolchains/XcodeDefault.xctoolchain/usr/bin/ranlib: archive member: .libs/libffi_convenience.a(ffi64.o) cputype (16777223) does not match previous archive members cputype (7) (all members must match)
/Applications/Xcode.app/Contents/Developer/Toolchains/XcodeDefault.xctoolchain/usr/bin/ranlib: archive member: .libs/libffi_convenience.a(unix64.o) cputype (16777223) does not match previous archive members cputype (7) (all members must match)
/Applications/Xcode.app/Contents/Developer/Toolchains/XcodeDefault.xctoolchain/usr/bin/ranlib: archive member: .libs/libffi_convenience.a(ffiw64.o) cputype (16777223) does not match previous archive members cputype (7) (all members must match)
/Applications/Xcode.app/Contents/Developer/Toolchains/XcodeDefault.xctoolchain/usr/bin/ranlib: archive member: .libs/libffi_convenience.a(win64.o) cputype (16777223) does not match previous archive members cputype (7) (all members must match)
libtool: link: ranlib .libs/libffi_convenience.a
/Applications/Xcode.app/Contents/Developer/Toolchains/XcodeDefault.xctoolchain/usr/bin/ranlib: archive member: .libs/libffi_convenience.a(ffi64.o) cputype (16777223) does not match previous archive members cputype (7) (all members must match)
[...]

PR #942, opened 2018-02-19
Drop Sass dependency to pre-3.5 to avoid ffi

johnfairh:
See #941 - I was hoping ffi would release a working version before too many people noticed this but it's been over a week now.

If you think it's worth doing a jazzy release in the interim then this patch pins ffi to the version jazzy has been using until a week ago ~cuts out the ffi dependency by back-levelling Sass to last summer~ -- doesn't change anything in the jazzy output.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,7 +10,9 @@
 
 ##### Bug Fixes
 
-* None.
+* Temporarily force ffi to 1.9.18 to avoid macOS build failures.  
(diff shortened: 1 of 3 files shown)


segiddins:
Is there any reason we can’t temporarily pin to an older ffi version?

johnfairh:
Only that I didn't think to try adding a new dependency directly from jazzy.  Yes that is much better.

johnfairh:
Not required any more.

PR #943, opened 2018-02-26
Add note about Xcode CLI tools to README

johnfairh:
Add a hint about Xcode CLI tools to installation instructions.


jpsim:
Thanks!

Issue #949, opened 2018-03-28
No current support for an "include" flag


NicholasFFox:
The `exclude` flag is awesome, but with our project's directory structure an `include` flag would make life a lot easier. This issue proposes adding an `include` flag that generates documentation *only* for included files.

Example usage:
```
jazzy --include my/project/src/public
```

johnfairh:
Fixed in master!

PR #950, opened 2018-03-28
Add support for an "include" option

NicholasFFox:
* Added support for -i or -include which has the opposite behavior of the exclude flag
 * Updated file filtering to account for include flag

 Addresses #949

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,7 +6,10 @@
 
 ##### Enhancements
 
-* None.
+* Added the ability to limit documentation to certain files by passing in an
+  `-i`/`--include` argument.  
+  [Nick Fox](https://github.com/nicholasffox)
+  [#949](https://github.com/realm/jazzy/issues/949)
 
 ##### Bug Fixes
 
--- a/lib/jazzy/sourcekitten.rb
+++ b/lib/jazzy/sourcekitten.rb
@@ -538,11 +538,33 @@ module Jazzy
       end
     end
 
+    # Apply filtering based on the "included" and "excluded" flags.
+    def self.filter_files(json)
+      json = filter_included_files(json) if Config.instance.included_files.any?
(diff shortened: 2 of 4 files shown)


NicholasFFox:
This implementation is used internally by our team. Opening this PR to facilitate discussion.

CHANGELOG.md
```diff
@@ -5,8 +5,11 @@
 * None.
 
 ##### Enhancements
-
-* None.
+--------------------------------------------------------------------------------
```
johnfairh: Don't need this line, see style in the rest of the file.  Could drop the 'to jazzy' part too, I think the context is clear.

lib/jazzy/sourcekitten.rb
```diff
@@ -538,11 +538,33 @@ def self.mark_members_from_protocol_extension(extensions)
...
+        doc[key]
+      end.compact
+    end
+
+    # Filter based on the "included" flag.
+    def self.filter_included_files(json)
+      included_files = Config.instance.included_files
+      json.map do |doc|
+        key = doc.keys.first
+        doc if included_files.detect do |foo|
```
johnfairh: Maybe a better variable name :-)

NicholasFFox: Oops! 🤦‍♂️

CHANGELOG.md
NicholasFFox: ✅

lib/jazzy/sourcekitten.rb
NicholasFFox: ✅

NicholasFFox:
@johnfairh I've addressed your feedback and updated the README. Let me know what you think!

johnfairh:
Great - thank you.

jpsim:
🎉

PR #951, opened 2018-03-29
Fix Dash docset index creation

cyphactor:
Prior to this Overview items were being represented in the Dash docset
index with a NULL type. Dash doesn't support indexes with NULL types in
them. To correct this I added an 'Overview' kind to the kind-type map
with a dash type of 'Section'. I talked about using this type with
Kapeli (the Dash developer) here,
https://github.com/Kapeli/Dash-User-Contributions/pull/1815 to see
if 'Section' was the appropriate type and it was.

This also includes new commit reference for the integration_specs which
I adjusted for this change as well. That pull request is open here,
https://github.com/realm/jazzy-integration-specs/pull/47

### DIFF
--- a/lib/jazzy/source_declaration/type.rb
+++ b/lib/jazzy/source_declaration/type.rb
@@ -5,7 +5,7 @@ module Jazzy
     # rubocop:disable Metrics/ClassLength
     class Type
       def self.all
-        TYPES.keys.reject { |k| k == 'Overview' }.map { |k| new(k) }
+        TYPES.reject { |_, v| v[:jazzy].nil? }.map { |k, _| new(k) }
       end
 
       attr_reader :kind
--- a/lib/jazzy/source_declaration/type.rb
+++ b/lib/jazzy/source_declaration/type.rb
@@ -5,7 +5,7 @@ module Jazzy
     # rubocop:disable Metrics/ClassLength
     class Type
       def self.all
-        TYPES.keys.map { |k| new(k) }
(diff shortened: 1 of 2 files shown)


lib/jazzy/source_declaration/type.rb
```diff
@@ -5,7 +5,7 @@ class SourceDeclaration
     # rubocop:disable Metrics/ClassLength
     class Type
       def self.all
-        TYPES.keys.map { |k| new(k) }
+        TYPES.keys.reject { |k| k == 'Overview' }.map { |k| new(k) }
```
johnfairh: How about doing the map as before, then rejecting `Type`s with nil name?  Reduces number of times 'Overview' is hard-coded + bit more future proof.

johnfairh:
Thanks a lot for the patch - good to see someone paying attention to the docset.

The specs changes look fine - I think you're right, the Dash anchor sections that have appeared were a casualty of the null type and should have been there all along.

One code comment, could you also add the bug fix to the changelog, see https://github.com/realm/jazzy/blob/master/CONTRIBUTING.md.

lib/jazzy/source_declaration/type.rb
cyphactor: Sounds good

cyphactor:
@johnfairh I have made all the suggested changes. Please let me know if there is anything else.

johnfairh:
👍 - merged via #953.

PR #954, opened 2018-04-04
Build specs with Xcode9.3/Swift 4.1

johnfairh:
Switch to Xcode93/Swift 4.1 seeing as Xcode94 beta is here....
* Lots of name mangling/USR-naming changes;
* SourceKit[ten] now pick up all `subscript` and `typealias` declarations even without doc comments;
* Some @available attributes showing up that didn't before.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -23,7 +23,7 @@ jobs:
 
   swift:
     macos:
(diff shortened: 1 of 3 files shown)


jpsim:
Niiiiiiiiiice.

PR #956, opened 2018-04-21
 Update SourceKitten to 0.20.0

jpsim:
No jazzy relevant changes in this SourceKitten release, other than maybe a minor performance improvement.

Also remove `install_name_tool` command no longer needed with Swift 4.1

### DIFF
--- a/Rakefile
+++ b/Rakefile
@@ -106,9 +106,6 @@ begin
       `swift build -c release -Xswiftc -static-stdlib`
     end
     FileUtils.cp_r "#{sk_dir}/.build/release/sourcekitten", 'bin'
-    system 'install_name_tool', '-delete_rpath',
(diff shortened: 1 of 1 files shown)


jpsim:
I lied about there being no jazzy relevant changes. Looks like we won improved sorting with SourceKitten 0.20.0, although I can't identify which change caused this: jpsim/SourceKitten@c3453d1...b6b1041

johnfairh:
I reckon https://github.com/jpsim/SourceKitten/commit/4005b43ddeec7c901793e97fcc5f85eac7530a42.
Means xcodeproject file list order doesn't affect declaration order.  Probably right to keep it consistent with SPM.

jpsim:
Right, that must be it. Thanks for finding that!

PR #960, opened 2018-04-27
Fix crash when specifying empty Swift version & update specs

jpsim:


jpsim:
I don't know _why_ specifying a Swift version of 4.1 leads to jazzy not finding a suitable Swift version, but jazzy's Swift version resolver has always been buggy.

Issue #447, opened 2016-01-03
Support mixed Objective-C and Swift projects


FelixSFD:
Im running jazzy (0.5.0) with the following command, but only my Swift-files are parsed. All Objective-C headers are ignored.

> jazzy --clean --objc --umbrella-header MyApp/MyApp-Bridging-Header.h --framework-root /Applications/Xcode.app/Contents/Developer/Platforms/iPhoneOS.platform/Developer/SDKs/iPhoneOS9.1.sdk --author "The Author" --sdk iphoneos --swift-version 2.1 --min-acl private --xcodebuild-arguments -scheme,MyAppScheme --module MyApp

What am I doing wrong?

khoogheem:
try it with
~~~
--framework-root . (assuming you are running jazzy in the root of your project)
~~~

Also, you shouldn't need --min-acl as that is for swift only I believe

FelixSFD:
If the root ist the folder, where all my source-files are ([folder where the .xcodeproj is located]/MyApp/, I'm in root.

But it didn't work. Still parsing my Swift files only.

I have Swift and Objective-C in my project, so I need it.

jpsim:
Sorry @FelixSFD, right now jazzy just documents *either* Objective-C or Swift projects. Generating documentation for Objective-C *and* Swift in the same project hasn't been added yet.

To my surprise, we didn't already have a ticket tracking that, so we're now using this one to track adding mixed documentation support.

eflath:
As a workaround, is it possible to run two different commands and create separate documentation for iOS and swift? the command I'm running now documents the Swift code only even though i'm passing the --objc flag and I can't get it to parse the objc files

jpsim:
Yes, you could combine the [SourceKitten](https://github.com/jpsim/SourceKitten) doc output for both Objective-C and Swift versions of your documentation using the `jazzy --sourcekitten-sourcefile`, but you're in *very* undocumented territory here so my ability to help you along will be limited.

ronak2121:
Any update on this? It's been almost 8 - 9 months on it now.

jpsim:
Nothing new to share, I don't think anyone's actively working on this. Would you like to contribute to this @ronak2121?

jpsim:
still no

wilmarvh:
Gonna take a guess here that there's no update yet @jpsim ?

If I were interested in contributing, where is the best place to start with this?

johnfairh:
[...]

PR #997, opened 2018-07-29
Minimize the amount of specs being built by jazzy and improve checking which targets to build.

### LINKED ISSUE #981: Pod subspecs are built (and documented) in addition to the main spec.
When one tries to run jazzy with a podspec that has multiple subspecs, jazzy will document all of them. This not only causes a lot of unnecessary compilation, but also get's jazzy confused, since a lot of symbols are defined multiple times.

Fix should be pretty simple here:
https://github.com/realm/jazzy/blob/83cebe59a40f5926662a5a6294ca1ba980149c5b/lib/jazzy/podspec_documenter.rb#L24-L27

Before I am opening a pr, should this be a new config option? If yes, where would I add those? (The first time I have worked with ruby, so assistance is appreciated).

I think something like: --subspecs Subspec and only include those targets.

Also I noticed that my own podspec creates some target with a weird label:

```
#<Pod::PodTarget:0x00007f908dcb9e70>
Pod-b2cd66da
Pod
#<Pod::PodTarget:0x00007f908dcb9808>
Pod-SocketServer
Pod
#<Pod::PodTarget:0x00007f908dcb92b8>
Pod-Image
Pod
```
Above was created by adding the following code:
```
installer.pod_targets.map do |t|
            warn t
            warn t.label
            warn t.pod_name
end
```

galli-leo:
This PR aims to improve jazzy's podspec integration in a few ways:

1. Jazzy will try to build specs as few times as possible. By keeping track of what specs are already built by previous targets, it can skip targets that only include specs of other targets (Mostly subspecs). If there are subspecs that have a dependency on other subspecs, this can still lead to jazzy running for a spec multiple times. However, that should be relatively rare and only happen when the default_subspecs are set.
2. Instead of checking the pod_name of a target and seeing if it matches, jazzy will take the first spec of a target and see whether the root of that target is the root of the podspec.

Let me know, if and where to add some tests for this as well as whether I should add a config option to only document certain subspecs.

Also should I refactor this? Rubocop complains that the method is too long.

Fixes #981

### DIFF
--- a/lib/jazzy/podspec_documenter.rb
+++ b/lib/jazzy/podspec_documenter.rb
@@ -21,8 +21,13 @@ module Jazzy
         installer = Pod::Installer.new(sandbox, podfile)
         installer.install!
         stdout = Dir.chdir(sandbox.root) do
+          specs_being_built = []
           targets = installer.pod_targets
-                             .select { |pt| pt.pod_name == podspec.root.name }
+                             .select do |pt|
+                               select = pt.specs[0].root == podspec.root && pt.specs.any? { |spec| !specs_being_built.include?(spec) }
+                               specs_being_built.push(*pt.specs)
+                               select
+                             end
                              .map(&:label)
 
           targets.map do |t|


galli-leo:
I just found an even better way of fixing this:

Instead of creating a target for every subspec and platform, we create a single target for each platform including all subspecs. This way, jazzy only has to run one xcodebuild per platform.

At the moment, I have left in the earlier code change. Let me know, if I should remove that and have jazzy run for every platform.

galli-leo:
Bump

joesus:
Is this still active? This would be useful to me.

johnfairh:
@joesus i think this is becalmed because none of the active maintainers are confident enough about the cocoapods internals to understand whether we really need a fix here, whether this is the right one, and what the implications are for existing users.

joesus:
@johnfairh that makes sense. I'm working on a combined Swift/ObjC project and still struggling to figure out the best approach for merging the documentation. The best I can come up with is to either 
a) Use sourcekitten directly for each language and then try to run jazzy directly against the files
b) Use jazzy to generate both and write a script to merge the files and use something like XSLT to combine the html.

In your opinion do you think option a will work or should I not waste time and start in on the more onerous option b?

johnfairh:
Probably (b) -- jazzy has an all-or-nothing approach to ObjC/Swift, more discussion in #447.

Issue #843, opened 2017-07-03
Parse `__attribute__((deprecated("Message")))` and show it in the documentation


michaelochs:
Objective-C methods marked as deprecated are currently not shown in the documentation generated by Jazzy as such. Everything needed for documenting this properly is already in the source code, even if there is no `@deprecated` attribute in the documentation, so Jazzy should parse this and make the user browsing the documentation aware of the fact that a method is deprecated.

Ideally Jazzy would show a yellow~ish box with a "Deprecated" title and the message from the deprecated attribute and also strike through the method name in the list. The latter would be very great but I would consider that as optional. It would already be a huge benefit for a developer to see that a method is deprecated when actually reading its documentation.

Another addition would be that if a `@deprecated` keyword is available in the comment for the methods, this should probably take precedence over the deprecated attribute in the source as the `@deprecated` keyword is explicitly there for documentation purpose, so I'd say this reasoning is valid.

When a complete class is marked as deprecated, the same box should be visible at the class documentation.

steipete:
This also seems mentioned in #541 and #368 however there was no dedicated issue so far.

mervynokm:
any chance of @deprecated being added in upcoming versions?
Would be great to support all the latest objective-c keywords

johnfairh:
The Objective-C sourcekitten json output already contains keys for deprecated/unavailable and any message from `__attribute__`, so someone could go ahead and do the jazzy front-end changes there.  Getting equivalent function for Swift needs some sourcekitten additions.

Supporting the headerdoc keyword is a bit different but is plausible once the presentation part is done.

meleyal:
> The Objective-C sourcekitten json output already contains keys for deprecated/unavailable and any message from `__attribute__`

Currently it looks like the deprecated/unavailable keys (`key.deprecation_message` / `key.unavailable_message`) in the SourceKitten json output are not populated, i.e. they don't get the message from `__attribute__`. Tested with both: 

```objc
__attribute__((deprecated("message"))
// and 
__deprecated_msg("message")
```

So I think for this to work SourceKitten would need some additions for both Objective-C and Swift.

Related: https://github.com/jpsim/SourceKitten/issues/214

johnfairh:
Closing this one because the ObjC part is done - can use good old #11 to track the Swift part for now.

PR #998, opened 2018-08-10
Mark unavailable and deprecated symbols

steviki:
Fixes https://github.com/realm/jazzy/issues/843.

This PR adds the ability to detect unavailable and deprecated symbols, as well as the related messages.

In the generated documentation, an unavailable or deprecated symbol is marked with a strikethrough of its name, and adds a box with the unavailable message or the deprecated message when expanding the symbol.

We are already using this in production for the PSPDFKit for iOS API documentation. (See https://pspdfkit.com/api/ios/Classes/PSPDFDocumentViewLayout.html#/c:objc(cs)PSPDFDocumentViewLayout(py)pagingEnabled for an example of a deprecated API.)

screen shot 2018-08-16 at 09 07 34

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,7 +6,9 @@
 
 ##### Enhancements
 
-* None.
+* Shows unavailable and deprecated symbols in the generated documentation via a strikethrough text attribute.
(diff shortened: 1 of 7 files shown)


johnfairh:
Continuing in #1041.

PR #1002, opened 2018-09-10
Correct docs about usage of the `readme_path` option.  The option string is actually `readme`.

joejune:


jpsim:
Thanks for the fix!

PR #1013, opened 2018-10-04
Fixed Apple theme for multiline copyright string

FabienLydoire:
Fixed the CSS to display properly a multiline copyright string without overlapping content.
Copyright text was overlapping documentation.


jpsim:
This also would appear to reduce the bottom padding for single-line copyright strings, which could be an issue.

johnfairh:
Thanks for this, will merge via #1039 to sort out the vertical spacing.

PR #1018, opened 2018-10-24
Update jazzy.search.js

karagiorgosm:
I added only 2 lines for the search functionality. Currently search works only for the name property. But it can work with the abstract property too.


jpsim:
Continued in #1027. Thanks for the PR!

PR #1019, opened 2018-10-30
Fixed Swift major version handling

erdnussflips:


erdnussflips:
Why this code changes breaks the tests? Any ideas?

johnfairh:
(you can see the test failure log by following the 'Details' link in the CI report)

The problem is that '4.0' is not a good default for `--swift-version`.  This needs to stay at `nil` to mean *whatever is configured as default for this shell*.

I think the right fix here is solely to `podspec_documenter` and needs to pass through a specified Swift version.  I *think* the "clip to major.0" logic comes from the Xcode 8 era when even though we were in Swift 3.2, the software required '3.0' to be passed here.  Since at least Swift 4.2 this checking has got more permissive.  So, taking it out would *probably* only break Xcode 8 users.  Maybe make it conditional on version major < 4?

johnfairh:
(working on fixing this up)

johnfairh:
Continued in #1037.

PR #1035, opened 2019-01-05
Add some installation problems/solutions

johnfairh:
Write up the space-in-Xcode-path problem.


jpsim:
Thanks for taking the time to do this 🙏

PR #1037, opened 2019-01-12
Support Swift 4.2 with --podspec

johnfairh:
Generalize the code that figures out SWIFT_VERSION from the full swift version number to work properly with 4.2 and future maj/min values.  Update the default version to 4.2 for Xcode 10.

Can't see how to avoid hard-coding these numbers.

Write up what the `--swift-version` flag actually does (most of the added lines.)

### DIFF
--- a/README.md
+++ b/README.md
@@ -219,6 +219,20 @@ Note that the `--include` option is applied before the `--exclude` option. For e
 Declarations with a documentation comment containing `:nodoc:` are excluded from the
 documentation.
 
+### Choosing the Swift language version
+
+Jazzy normally uses the Swift compiler from the Xcode currently configured by
+`xcode-select`.  Use the `--swift-version` flag to compile with a different
+Xcode.
+
+The value you pass to `--swift-version` must be the Swift language version given
+by `swift --version` in the Xcode you want to use.
+
+For example to use Xcode 9.4:
+```shell
+jazzy --swift-version 4.1.2
+```
+
 ## Troubleshooting
 
 #### Swift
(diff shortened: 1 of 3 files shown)


README.md
```diff
@@ -227,6 +240,16 @@ documentation.
...
 
+**Unable to find an Xcode with swift version X**
+
+1. The value passed with `--swift-version` must exactly match the version
+   number from `swiftc --version`.  For example Xcode 10.1 needs
+   `--swift-version 4.2.1`.  See [the flag documentation](#choosing-the-swift-language-version).
+2. The Xcode you want to use must be in the Spotlight index.  You can check
+   this using `mdfind 'kMDItemCFBundleIdentifier == com.apple.dt.Xcode'`.
+   Some users have reported this issue being fixed by a reboot; `mdutil -E`
+   may also help.
```
jpsim: May want to say that if the Xcode isn't tin the Spotlight index, another alternative is to specify it via `DEVELOPER_DIR`

lib/jazzy/podspec_documenter.rb
```diff
@@ -97,6 +97,23 @@ def self.github_file_prefix(podspec)
 
     private_class_method :github_file_prefix
 
+    # Latest valid value for SWIFT_VERSION.
+    LATEST_SWIFT_VERSION = '4.2'.freeze
```
jpsim: Should this be `4.2.1`?

lib/jazzy/podspec_documenter.rb
```diff
@@ -97,6 +97,23 @@ def self.github_file_prefix(podspec)
...
+    # All valid values for SWIFT_VERSION that are longer
+    # than a major version number.  Ordered ascending.
+    LONG_SWIFT_VERSIONS = ['4.2'].freeze
+
+    # Go from a full Swift version like 4.2.1 to
+    # something valid for SWIFT_VERSION.
+    def compiler_swift_version(user_version)
+      return LATEST_SWIFT_VERSION unless user_version
+
+      LONG_SWIFT_VERSIONS.select do |version|
```
segiddins: `LONG_SWIFT_VERSIONS.find ...` ?

README.md
```diff
@@ -219,6 +219,19 @@ Note that the `--include` option is applied before the `--exclude` option. For e
 Declarations with a documentation comment containing `:nodoc:` are excluded from the
 documentation.
 
+### Choosing the Swift language version
+
+Jazzy normally uses the Swift compiler from the currently selected Xcode.  Use
```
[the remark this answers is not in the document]
johnfairh: OK

README.md
johnfairh: OK

lib/jazzy/podspec_documenter.rb
johnfairh: Nope, here we are working out values to pass to `swiftc -swift-version` via `xcodebuild`'s `SWIFT_VERSION`.  `4.2.1.` isn't valid for `swiftc` and if you pass it to `xcodebuild` it gets turned into `4`.

Issue #552, opened 2016-05-04
Question about 'Other ' prefix with custom_categories


lumaxis:
I've been playing with the `custom_categories` option in jazzy a bit yesterday and was a bit surprised that it adds the "Other " prefix to _all_ categories whenever any custom category is used. From a gut feeling, I would have expected the prefix only be added to duplicate categories, i.e. when `custom_categories` also contains "Classes" for example.
Can someone shed some light on the reasoning behind this? Personally, I would vote for this feature to be either completely removed or only applied to duplicate categories. Depending on what the opinions here are, I'm happy to apply any changes if necessary :)
Thanks!

These are the code parts that add the "Other: " prefix:
https://github.com/realm/jazzy/blob/927079bc2b538eca8725eb13c3544b3146a28fc0/lib/jazzy/sourcekitten.rb#L24
https://github.com/realm/jazzy/blob/927079bc2b538eca8725eb13c3544b3146a28fc0/lib/jazzy/sourcekitten.rb#L50

istx25:
Hi @lumaxis. Thanks for filing this issue. My apologies that no one has responded to you until now. I totally see where you're coming from. My personal opinion is that I agree with you. It definitely makes sense to only add Other on duplicated categories. With that said, I'd like to rope in @jpsim, @segiddins and @orta to hear they have to say. And PRs are always welcome.

SteffenL:
Hello!

Since this issue has been open for quite some time now without further discussion, I wonder if this is still a valid issue that you are willing to discuss further.

Personally, I cannot see any reasons why I would want a section named "Other ..." without explicitly naming it so myself.

I have just started experimenting with jazzy and custom categories myself. I only wanted to add a "Documentation" section with links to handwritten documentation pages. Now all of the other sections (Classes, Protocols, etc) have been prefixed with "Other", making it look quite strange.

lumaxis:
Personally, I have moved on from working on iOS things, so this is not directly relevant to me anymore. Since this also doesn't seem to be relevant for anyone else, closing.

EdgarMaucourant:
Hi,
In case someone find this post and wonder what could be the a solution, there is actually an undocumented attribute you can use in your config file to define the value of this prefix (default to Other) set it to "" to get rid of it. Attribute name is "custom_categories_unlisted_prefix"

Edgar

PR #1047, opened 2019-02-13
"Other" category prefix

SteffenL:
Since there has not been any discussion in issue #552 for quite some time, I am making this obviously incomplete PR to encourage that we discuss this again. I can finish the rest when we decide exactly what to do about this issue.

Related PR: https://github.com/realm/jazzy-integration-specs/pull/78

Please note that I have just started using jazzy and I am unfamiliar with how it should work, potential side effects, etc.

Removing the "Other" category prefix works best for me both when using custom categories and when not using them. I do not see any reasons why I would name any custom categories so that I end up with duplicate names, and I do not see any reason why I would want "Other" in the name without explicitly naming a category that way myself.

Therefore I cannot currently see any reasons to keep this prefix, but I guess it is not as simple as just removing it completely, as I am sure there was a good reason for this.

What do you think?

Note: Uses a custom branch of `jazzy-integration-specs`.

### DIFF
--- a/lib/jazzy/sourcekitten.rb
+++ b/lib/jazzy/sourcekitten.rb
@@ -62,9 +62,7 @@ module Jazzy
     # Group root-level docs by custom categories (if any) and type
     def self.group_docs(docs)
       custom_categories, docs = group_custom_categories(docs)
-      type_categories, uncategorized = group_type_categories(
-        docs, custom_categories.any? ? 'Other ' : ''
-      )
+      type_categories, uncategorized = group_type_categories(docs)
       custom_categories + type_categories + uncategorized
     end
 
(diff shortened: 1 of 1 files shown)


johnfairh:
Thanks for the PR, nice job navigating the integration specs.

Firstly I don't understand what you mean about seeing 'Other Xxx' when _not_ using `custom_categories`.  That shouldn't happen; can you say more about the scenario?

Ok. The idea of `custom_categories` is that instead of grouping declarations by kind they are grouped by use case or some logical category that makes sense to the user of the documentation.

Lots of docsets I've seen cover all their decls in these categories.  Lots of others don't -- because they have categories for maybe 1 / 2 core use cases only -- and end up with a bunch of left-over declarations.  Jazzy groups these into the 'Other Xxx' headings to imply "hey here are a bunch of classes, but btw not _all_ classes in the docset are here, just those not covered by the categories above".

So I don't think just deleting all mention of 'Other' is a good plan.  The approach sketched in #552 still feels logical: only use 'Other' in a category heading if at least one top-level decl of that kind has been included by custom categories.

SteffenL:
Thanks for your response!

> Firstly I don't understand what you mean about seeing 'Other Xxx' when _not_ using `custom_categories`. That shouldn't happen; can you say more about the scenario?

Sorry, I just mean that it is more consistent for my use case. As soon as I add a category for handwritten documents, e.g. "Documentation", it was unexpected that all other type categories showed up as "Other ..." since I did not group any types under that "Documentation" category.

I think I see why there is a need for an "Other ..." prefix now when grouping types in custom categories, and the other suggested solution sounds better to me now.

I will think a bit more about whether I can take on that job (sounds easy in theory) and perhaps update the PR.

PR #1048, opened 2019-02-13
Update javascript dependencies

johnfairh:
Update jQuery, Lunr, typeahead to latest releases.  Search finds maybe a few more things?  I've enabled the new option to let you hit 'enter' to select the top search hit.

Specs updates to include new js files.

Also remove workaround #714 fixed by #628.

Not planning to changelog this.


segiddins:
Probably worth changelogging in case someone cares about dependency versions for security reasons?

johnfairh:
Yes OK, makes sense.

jpsim:
Nice! Thanks so much for all your continued maintenance of this project @johnfairh! 🙏

Issue #788, opened 2017-04-26
Clicking a linked member should unfold its documentation


reitzig:
Clicking on a member of some type, I am taken to the corresponding site and the member documentation is unfolded.

Not so if the member is on the current site: then, the browser (FF 52.0.2) jumps to the anchor but does not unfold the documentation. It should do just that.

johnfairh:
Fixed in master.

PR #1051, opened 2019-02-27
Fix section open when linked to from current page

johnfairh:
This is a javascript fix to navigation involving the openable/closable sections.

Main bug fixed is #788:
* Click to open section A.
* Click a link to section B on the same page.
* Before: scroll to B but B does not open; after: B opens.
   
Subtly different bug:
* Click to open section A.
* Click to close section A.
* Click a link to A.
* Before: scroll to A but nothing happens; after: A opens.
   
Also move a couple of things to CSS so that the `jazzy.js` file is 100% identical between the themes: 'changed lines' stat inflated by 3 copies of this.

Spec changes to reflect shipped js/css files.  Manually verified dash is still fine.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,7 +10,9 @@
 
 ##### Bug Fixes
 
-* None.
+* Unfold member documentation when linked to from current web page.  
+  [John Fairhurst](https://github.com/johnfairh)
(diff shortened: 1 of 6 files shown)


johnfairh:
Thanks; yes will do - wanted another pair of eyes on my naive javascript.

PR #1058, opened 2019-03-29
Swift 5 updates

johnfairh:
Version number updates and a couple of compiler bug/feature workarounds.

SourceKit's calculation of accessibility has changed in Swift 5: internally it
now uses what the core compiler calculates rather than making it up
independently.  This may have knock-ons for some docsets.

Xcode 10.2 spec changes:

* Alamofire - nothing.
* Moya - some changes to printing of generic decls.
* Realm-ObjC - Decl printing changes, libclang updates in this Xcode!
* Realm-Swift - mostly somewhat ugly churn triggered by compiler changes in how SourceKit handles `#if` conditional compilation.  Will need to do something thoughtful about this but not right now.  Some generic decl printing changes.
* Siesta - whitespace decl-printing change.
* MiscJazzyFeatures - stdlib change knock-ons for an extension.
* MiscJazzyObjCFeatures - nothing.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -23,7 +23,7 @@ jobs:
 
   swift:
     macos:
-      xcode: "10.0.0"
+      xcode: "10.2.0"
     environment: *bundler-environment
     # Used to invoke chruby
     shell: *shell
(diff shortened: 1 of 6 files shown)


jpsim:
Thanks so much!

Issue #1057, opened 2019-03-28
Should upgrade Sourcekitten to an Xcode 10.2 compatible version


FabienLydoire:
Jazzy fails on workspace doc generation using Xcode 10.2 and Swift 5.0.
Seems like it does not find the project (or at least the Signing details).

```
Build settings from command line:
    CODE_SIGN_IDENTITY = 
    CODE_SIGNING_REQUIRED = NO

note: Using new build system

** CLEAN SUCCEEDED **

note: Using new build system
note: Planning build
note: Constructing build description
Build system information
error: An empty identity is not valid when signing a binary for the product type 'Application'. (in target '<Edited>')

** BUILD FAILED **
```

Building Sourcekitten from the master branch (commit 2b5d4448b67a9de790f9fbb02cbac92e5a92060d) and embedding the newly build version into Jazzy (replacing `/Library/Ruby/Gems/2.3.0/gems/jazzy-0.9.5/bin/sourcekitten` ) fixes the issue.

johnfairh:
Thanks; will be updating sourcekitten in the next few days.

However, that is not the underlying problem: latest sourcekitten still passes `CODE_SIGN_IDENTITY=` when it has to do `xcodebuild clean` -- it just usually manages to avoid doing that kind of build (by parsing the llbuild manifest).

This error is new in Xcode 10.2 only with the new build system; playing around with a signed project it looks like adding `CODE_SIGNING_ALLOWED=NO` does the trick and is harmless on other configs.

FabienLydoire:
Thanks @johnfairh 
Would you mind describing how you can add `CODE_SIGNING_ALLOWED=NO` in the Jazzy configuration file ?

johnfairh:
You could try `jazzy -x CODE_SIGNING_ALLOWED=NO`; I was looking at `xcodebuild` itself.

FabienLydoire:
`xcodebuild_arguments: [CODE_SIGNING_ALLOWED=NO]` worked.

Thanks @johnfairh !

JavieriOSninja:
It is still not working for me, this is the output that I receive:

Using config file /Users/ios/Documents/iOS_Projects/SecretSanta/App/Code/.jazzy.yaml
Running xcodebuild
Could not parse compiler arguments from `xcodebuild` output.
Please confirm that `xcodebuild` is building a Swift module.
[...]

PR #1063, opened 2019-04-04
Update dependencies and release 0.9.6

johnfairh:
Sourcekitten to 0.22 for various fixes in particular most of #1057.
Tiny specs update because latest RxSwift no longer pulls in RxAtomic...


jpsim:
🎉 thanks @johnfairh !

Issue #810, opened 2017-05-11
Set version of generated Docset


reitzig:
If I import the generated Docset into "Docs for Xcode", I see `Version (null) (null)`. `module_version` is set in the Jazzy config file.

jpsim:
Xcode docsets aren't generated... are you using Dash? Can you share more information?

reitzig:
I don't use Dash (what is that?), but a Docset still appears in `<build>/docsets`.

I call Jazzy with options `--clean`, `--config <my file>` and `--sdk iphoneos`. In the config file, I set the output directory, custom categories, documentation, readme and abstract, plus some meta data.

jpsim:
The Docset is a Dash-format docset: https://kapeli.com/dash

Jazzy doesn't generate Xcode-formatted docsets.

You say you see `Version (null) (null)`, but you don't say where you see that... 🤔

reitzig:
Ah, okay. I tried opening it with XCode or whatever tool macOS has by default. If that's not supposed to work, this ticket is void.

Issue #666, opened 2016-09-29
Module version is not available in templates


rlepinski:
The module version is unavailable in templates:

```
{{module_version}}
```

Doesn't look like its set when generating the doc.

istx25:
What version?

istx25:
Please share a minimal test project that reproduces the issue.

istx25:
Closing in 2 weeks if there's no further activity.

jpsim:
None of the templates have `{{module_version}}`. @rlepinski is this a feature request? Would you like jazzy to make the module version available to Mustache templates?

rlepinski:
@istx25 version 0.7.2.

@jpsim yes feature request. None of the templates use it but the command line options accept the module version so it would be nice to expose it for moustache templates.  Sorry for not providing more information when I originally created the ticket.

jpsim:
Sure, that's easily exposable. Let's just add it to [doc.rb](https://github.com/realm/jazzy/blob/master/lib/jazzy/doc.rb) much like `language` is currently exposed.

johnfairh:
Fixed in master.

PR #1065, opened 2019-04-18
Mustache templates can now expose `{{module_version}}`

mbalex99:
- I added module_version to expose `Config.instance.version`
- This can now expose `{{module_version}}` in the mustache templates. 

#810 #666 ##1064


mbalex99:
Only thing is I'm not sure where the html comparison code is to help fix the tests that are failing.

johnfairh:
Thanks for this; continued in #1077.

PR #1070, opened 2019-05-07
Updates for Ruby sass -> sassc gem

johnfairh:
This PR swaps out the `Sass` gem for `SassC` because the former is end of life and causes lengthy warnings on installation.  `SassC` is a wrapper of the C implementation and has more features etc. which is good - the .css produced is identical bar a handful of newlines.

No great rush to do this: current gem works fine, and we'd ideally wait for a `SassC` release including  sass/sassc-ruby#96 to improve the native build.  This also changes the min Ruby from 2.0 to 2.3, which should be fine: 2.3 itself has just gone eol.  High Sierra has system Ruby 2.2 though, so anyone still there will need to upgrade something.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,7 +6,9 @@
 
 ##### Enhancements
 
-* None.
+* Sass support now provided by `libsass` via `sassc` instead of the
+  deprecated Ruby Sass gem.  
+  [John Fairhurst](https://github.com/johnfairh)
 
 ##### Bug Fixes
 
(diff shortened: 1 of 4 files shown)


jpsim:
Looks great! Merge when you have a compelling reason to do so 😄

johnfairh:
Very close now - sassc has made it to 2.1.0.pre3...

johnfairh:
Merging with sassc 2.1.0.  They've removed the Ruby 2.3 requirement by popular demand which makes life easier.

PR #1073, opened 2019-05-17
Support CocoaPods 1.6+

johnfairh:
Make it work, and also respect `swift_version[s]` in the podspec.

The interesting `podspec_documenter` change is to set the target swift version before `Pod::Installer.install` which CP 1.6+ needs to avoid _does not specify a Swift version and none of the targets integrating it have the `SWIFT_VERSION` attribute set_ fatal errors.  @segiddins could you check this over / recommend a better way please?  Code inspired by / copied from `validator.rb`, tested on 1.5/1.6/1.7 rc2.

@jpsim I've updated the gemspec for only CP to be `~> maj.min`.  We used to do this for all the dependencies but changed to `~> maj.min.patch` [last Nov](https://github.com/realm/jazzy/commit/4414536e879366a963c7cd540a165904f36064b9#diff-ec08b1c2a5d1a60ba1201becd7fa168f) --- was that a slip or the way you'd still prefer it to be?


jpsim:
`maj.min` is fine. I don’t recall explicitly setting the patch, nor why I would have done that.

Thanks for the PR!

johnfairh:
OK great - will update the other deps in a separate one.

PR #1089, opened 2019-08-03
SPM integration

johnfairh:
This is a prototype set of changes to add first-class SPM support.

The goal is to let you type just `jazzy` and have that make docs from a `Package.swift` if there's no Xcode project (or the platform isn't Darwin, eventually).

First commit here is real, second is a few temp hacks to build one of the test specs with SPM instead of Xcode.


lib/jazzy/config.rb
```diff
@@ -192,6 +192,20 @@ def expand_path(path)
...
+    config_attr :swift_build_tool,
+      command_line: "--swift-build-tool #{SWIFT_BUILD_TOOLS.join(' | ')}",
+      description: 'Control whether Jazzy uses Swift Package Manager or '\
+                   'xcodebuild to build the module to be documented.  By '\
+                   'default it uses xcodebuild if there is a .xcodeproj '\
+                   'file in the source directory.',
+      parse: ->(tool) do
+        return tool.to_sym if SWIFT_BUILD_TOOLS.include?(tool)
+        raise "Unsupported swift_build_tool #{tool}, "\
+              "supported values: #{SWIFT_BUILD_TOOLS}"
```
jpsim: How is this rendered? Should we `join(', ')`?

lib/jazzy/sourcekitten.rb
```diff
@@ -183,16 +183,25 @@ def self.rec_path(path)
       end.select { |x| x }.flatten(1)
     end
 
+    def self.use_spm?(options)
+      options.swift_build_tool == :spm ||
+        (!options.swift_build_tool_configured &&
+         Dir['*.xcodeproj'].empty?)
```
jpsim: Nice 💯

lib/jazzy/config.rb
johnfairh: Oh well spotted.  Ruby renders it OK but that's not what I meant to write!

johnfairh:
> Does this need more than jpsim/SourceKitten#611 

Yep, I need to do one more sourcekitten PR to run `swift build` there.  Next few days?

lib/jazzy/doc_builder.rb
```diff
@@ -55,21 +55,13 @@ def self.doc_structure_for_docs(docs)
...
-        if options.podspec_configured
-          pod_documenter = PodspecDocumenter.new(options.podspec)
-          stdout = pod_documenter.sourcekitten_output(options)
-        else
-          stdout = Dir.chdir(options.source_directory) do
-            arguments = SourceKitten.arguments_from_options(options)
-            SourceKitten.run_sourcekitten(arguments)
-          end
-        end
-        unless $?.success?
```
johnfairh: this `unless` block is unreachable, lower levels raise instead.

johnfairh:
OK I think this is done now, if you could take a look over the last couple of commits when you've a chance!

johnfairh:
>Linux support
Definitely - could at least look for sourcekitten in $PATH and/or have a flag to set it.

(windows might be a ways off...)

PR #1093, opened 2019-08-25
Preserve non-latin characters in guide filenames and heading IDs

### LINKED ISSUE #1091: Wrong parsing with chinese
Thanks for great project!

Jazzy just works fine with english characters but I found that it has some problem when using unicode characters (like Chinese.) in jazzy.json like below:

`
"theme": "fullwidth",
    "documentation": "doc/MarkDown/*.md",
"custom_categories": [{
        "name": "简介",
        "children": [
            "名词说明",
            "功能介绍",
            "工程配置",
        ]
    }]
`

Jazzy can’t  generate  `名词说明.html` that match the dictionary key ‘children’

johnfairh:
Fixes #1091.

Jazzy is too aggressive at stripping out 'bad' characters when making 'friendly' identifiers.  This is a simple change that uses a unicode-friendly 'letter' definition.  We did this in another place too - when generating href IDs for markdown headings.

Added a demo of these to the sample project.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,7 +12,9 @@
 
 ##### Bug Fixes
 
-* None.
+* Preserve non-latin characters in guide filenames and heading IDs.  
+  [John Fairhurst](https://github.com/johnfairh)
+  [#1091](https://github.com/realm/jazzy/issues/1091)
 
 ## 0.10.0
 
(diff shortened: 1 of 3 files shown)


johnfairh:
Merging this one to make the next part easier.

jpsim:
Nice!

PR #1106, opened 2019-09-17
Update version to 0.11.1

johnfairh:
Release 0.11.1 to fix the xcodebuild breakage.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,4 +1,4 @@
-## Master
+## 0.11.1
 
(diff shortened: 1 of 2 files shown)


jpsim:
Thanks for cutting the release! In the future, be sure to run `bundle install` after updating the version to reflect the change in `Gemfile.lock`.

johnfairh:
🤦‍♂ rats, I wanted to avoid updating the dependencies for this but missed jazzy itself was in there.

Issue #921, opened 2017-12-05
Doesn't generate proper docset without --module


hollanderbart:
With several internal projects I experience the issue where the docset simple isn't generated. The folders stays empty and I'm using the default build instruction. Can I get insight in some jazzy logging? `xcodebuild` runs fine and HTML looks fine also.

johnfairh:
Try `ls -a` in the `docs/docsets/` directory?  The only gotcha I can see is that if you do not give jazzy  `--module` then it uses an empty string for the docset name, accidentally creating hidden files.

I think we should skip docset creation + warn if the module name is not set: the name is used in various dash metadata as well as filenames so defaulting to 'Jazzy' or something doesn't feel right + won't point users towards what to fix.  I doubt anyone is currently relying on the oddly named docsets.

hollanderbart:
Providing the `-- module` or `-m` parameter solved it! Thanks!

johnfairh:
I'm going to keep this open to do some code changes to avoid/warn about this.

hollanderbart:
Great to see that you're going to improve the experience! 👍

PR #1113, opened 2019-09-28
Initial work for mixed Swift and ObjC Sourcekitten input

joesus:
Initial attempt at mixing Swift and ObjectiveC.

Todo:
* Unit tests
* A process / instructions for providing input
* Figuring out how to get language stub to appear in the template (following the same pattern as `language` did not  seem to work, probably missed a step)

### DIFF
--- a/lib/jazzy/config.rb
+++ b/lib/jazzy/config.rb
@@ -151,9 +151,8 @@ module Jazzy
       command_line: '--hide-declarations [objc|swift] ',
       description: 'Hide declarations in the specified language. Given that ' \
                    'generating Swift docs only generates Swift declarations, ' \
-                   'this is only really useful to display just the Swift ' \
-                   'declarations & names when generating docs for an ' \
-                   'Objective-C framework.',
+                   'this is useful for hiding a specific interface for either ' \
+                   'Objective-C or mixed Objective-C and Swift projects.',
       default: ''
 
     config_attr :config_file,
@@ -172,10 +171,10 @@ module Jazzy
       command_line: ['-x', '--xcodebuild-arguments arg1,arg2,…argN', Array],
       description: 'Back-compatibility alias for build_tool_arguments.'
 
-    config_attr :sourcekitten_sourcefile,
-      command_line: ['-s', '--sourcekitten-sourcefile filepath1,…filepathN',
+    config_attr :sourcekitten_sourcefiles,
+      command_line: ['-s', '--sourcekitten-sourcefiles filepath1,…filepathN',
                      Array],
-      description: 'Files generated from sourcekitten output to parse',
+      description: 'File(s) generated from sourcekitten output to parse',
       parse: ->(paths) { paths.map { |path| expand_path(path) } }
 
     config_attr :source_directory,
--- a/lib/jazzy/config.rb
+++ b/lib/jazzy/config.rb
@@ -165,9 +165,10 @@ module Jazzy
       description: 'Back-compatibility alias for build_tool_arguments.'
 
     config_attr :sourcekitten_sourcefile,
-      command_line: ['-s', '--sourcekitten-sourcefile FILEPATH'],
-      description: 'File generated from sourcekitten output to parse',
-      parse: ->(s) { expand_path(s) }
+      command_line: ['-s', '--sourcekitten-sourcefile filepath1,…filepathN',
+                     Array],
+      description: 'Files generated from sourcekitten output to parse',
+      parse: ->(paths) { paths.map { |path| expand_path(path) } }
 
     config_attr :source_directory,
       command_line: '--source-directory DIRPATH',
--- a/lib/jazzy/config.rb
+++ b/lib/jazzy/config.rb
@@ -102,6 +102,14 @@ module Jazzy
       Pathname(Dir[abs_path][0] || abs_path) # Use existing filesystem spelling
     end
 
+    def hide_swift?
+      hide_declarations == 'swift'
+    end
+
+    def hide_objc?
+      hide_declarations == 'objc'
+    end
+
     # ──────── Build ────────
 
     # rubocop:disable Layout/AlignParameters
--- a/lib/jazzy/config.rb
+++ b/lib/jazzy/config.rb
@@ -151,8 +151,9 @@ module Jazzy
       command_line: '--hide-declarations [objc|swift] ',
       description: 'Hide declarations in the specified language. Given that ' \
                    'generating Swift docs only generates Swift declarations, ' \
-                   'this is useful for hiding a specific interface for either ' \
-                   'Objective-C or mixed Objective-C and Swift projects.',
+                   'this is useful for hiding a specific interface for ' \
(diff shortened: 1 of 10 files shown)


joesus:
@johnfairh, this clearly needs quite a bit of work still but wanted to get some eyes on it early. Thank you for the help in getting this far!

lib/jazzy/source_declaration.rb
```diff
@@ -13,6 +14,20 @@ def render_as_page?
...
+    def language
+      type.swift_type? ? 'Swift' : 'Objective-C'
+    end
+
+    def swift
+      type.swift_type?
+    end
+
+    # Magic strings for use by the specific syntax highlighting library
+    # Might be better to have this mapping take place in Highlighter.rb
```
johnfairh: I agree, could have Highlighter::SWIFT etc. constants and `Highlighter.highlight_swift(source)` helpers to make the caller simpler.

lib/jazzy/sourcekitten.rb
```diff
@@ -296,14 +297,14 @@ def self.process_undocumented_token(doc, declaration)
       make_default_doc_info(declaration)
 
       filepath = doc['key.filepath']
-      objc = Config.instance.objc_mode
-      if objc || should_mark_undocumented(filepath)
+
+      if !declaration.swift? || should_mark_undocumented(filepath)
```
johnfairh: This `swift?` doesn't match the method name in `SourceDeclaration`.

lib/jazzy/sourcekitten.rb
```diff
@@ -319,17 +320,20 @@ def self.parameters(doc, discovered)
       end.reject { |param| param[:discussion].nil? }
     end
 
+    # rubocop:disable Metrics/MethodLength
     def self.make_doc_info(doc, declaration)
       return unless should_document?(doc)
 
-      if Config.instance.objc_mode
+      if declaration.swift?
```
johnfairh: This `swift?` doesn't match the method name, is probably why CI is failing.

lib/jazzy/sourcekitten.rb
```diff
@@ -319,17 +320,20 @@ def self.parameters(doc, discovered)
       end.reject { |param| param[:discussion].nil? }
     end
 
+    # rubocop:disable Metrics/MethodLength
```
johnfairh: Generally prefer to avoid turning this one off.

lib/jazzy/source_declaration.rb
```diff
@@ -13,6 +14,20 @@ def render_as_page?
       children.any?
     end
 
+    def language
```
johnfairh: worth a comment that this is a user-facing string.

johnfairh:
Added a commit to sort out the highlighting problems that were breaking CI -- feel free to throw it away + refactor as you prefer.  One breakage was the case where the stuff being hilighted is from markdown (eg. readme) so neither swift nor objc.  Not sure why it was sometimes getting the language backwards.

joesus:
Thanks for fixing the highlighting code! What's the next steps for this? CLI flag to pass in Jazzy input to combine? Or maybe just documenting on the sourcekitten command that it will take a file that has mixed jazzy input? What do you think would be least impact? (Assuming least impact is the goal for this change)

joesus:
@johnfairh I'll try and take this piece by piece:

1. Test and fix any breakage of --hide-declarations
**Fixed**

2. Add a Swift file to misc_jazzy_objc_features and update spec recipe to include mixed project in CI 
**Need to do still.**

3. Decide what to do about language_stub & do it
**Need to do still.**

4. Check what Dash does with a mixed docset
**Need to do still.**

5. Add usage instructions to readme
**Added**

Hopefully will have some time tomorrow to tackle item 2.
Thanks!

lib/jazzy/config.rb
```diff
@@ -164,10 +171,10 @@ def expand_path(path)
       command_line: ['-x', '--xcodebuild-arguments arg1,arg2,…argN', Array],
       description: 'Back-compatibility alias for build_tool_arguments.'
 
-    config_attr :sourcekitten_sourcefile,
-      command_line: ['-s', '--sourcekitten-sourcefile filepath1,…filepathN',
+    config_attr :sourcekitten_sourcefiles,
```
johnfairh: I didn't rename this earlier because it's not backwards compatible (in the config file).

lib/jazzy/sourcekitten.rb
```diff
@@ -267,6 +267,9 @@ def self.should_document?(doc)
       # Always document Objective-C declarations.
       return true unless type.swift_type?
 
+      # Don't document Swift types if we are hiding Swift
+      return false if type.swift_type? && Config.instance.hide_swift?
```
johnfairh: (already know `type.swift_type?` is `true` here)

joesus:
@johnfairh, sort of unclear how these changes broke the specs... did I need to make an updating commit to the specs repo?

joesus:
@johnfairh, update: 

> Check what Dash does with a mixed docset

Seems to work the same as the website. Notably it does generate the docset as a hidden file due to lack of a module name but it seems this is a known issue (#921). 

This is what the mixed docs looks like in Dash. (changed the docset name so that it could be imported)
image

lib/jazzy/sourcekitten.rb
```diff
@@ -307,7 +307,7 @@ def self.process_undocumented_token(doc, declaration)
         declaration.abstract = undocumented_abstract
       else
         declaration.abstract = Markdown.render(doc['key.doc.comment'] || '',
-                                               declaration.language)
+                                               declaration.display_language)
```
johnfairh: My bad - should have changed to the highlighter language in a previous.

johnfairh:
Ignore the cocoapods CI failure for now - CI issue to do with cocoapods setup, I think fixed on master.

joesus:
This is coming together nicely! I'll take a crack at updating the specs but might not get to it until this weekend. Any tips or pitfalls to watch out for? Thanks!

johnfairh:
Roughly: rebase this branch to master, fork and branch realm/jazzy-integration-specs and change the `spec/integration_specs` submodule to point there.  Update `misc_jazzy_objc_features`, then the driver code to invoke jazzy is in `spec/integration_spec.rb`.  `bundle exec rake rebuild_integration_fixtures` to generate new stuff to check in - shouldn't change any other projects.

johnfairh:
Updated the MiscJazzyObjCFeatures project to include a Swift file and some cross-language extensions to show how they (aren't) handled.

joesus:
This is great! I’ll start on the follow-up this weekend. :)

jpsim:
Nice. This feature has been years in the making, thanks for pushing it through @joesus & @johnfairh 🎉

bamx23:
When these changes are expected to be in release(0.11.3 or 0.12.0)?

johnfairh:
There's a bug we need to fix to do with contents pages being overwritten.  I should have some time next week if @joesus doesn't beat me to it.

PR #1127, opened 2019-11-06
Merge Swift and ObjC categories of the same kind

johnfairh:
For mixed Swift and ObjC projects, generate merged 'classes' pages instead of generating two separate pages with the same name.

Spec project updates show the change:

Screenshot 2019-11-06 at 12 32 01

I tried a couple of other clever data structure things but it turns out the order of the `TYPE` array determines the order in the site making it tough to not lose that, so, brute force.  The categories list is not very long.

@joesus can you take a look please?


joesus:
@johnfairh - Tested it out! Looks great! Thanks for doing this!

PR #1139, opened 2019-12-09
Release 0.13.0 and update gems

johnfairh:


jpsim:
Thanks for all the hard work that went into this release, @johnfairh! 👏

johnfairh:
Feels good to finally get these done!

PR #1149, opened 2020-01-07
Use best declaration with @escaping or @autoclosure

### LINKED ISSUE #1148: typealias for function signature on multiple lines ignores all but 1st line
Here's what we have in ReSwift, a typealias on 2 lines:

```swift
public typealias Middleware<State> = (@escaping DispatchFunction, @escaping () -> State?)
    -> (@escaping DispatchFunction) -> DispatchFunction
```

Here's what jazzy generates:

```
public typealias Middleware<State> = (@escaping DispatchFunction, @escaping () -> State?)
```

([See our docs](http://reswift.github.io/ReSwift/master/Other%20Typealiases.html#/s:7ReSwift10Middlewarea))

I tried this with a clean install of the latest 0.13.0 and the result is the same.

johnfairh:
Improve the SR-631 workaround (@escaping dropped from declaration) so it doesn't fire if the version of Swift in use has the fix.  Add a test.

Fixes #1148 (at least gets a correct declaration into docs, getting the preferred multi-line one is a separate problem...)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -31,6 +31,10 @@
 * Fix crash when SourceKit returns out of bounds string byte offsets.  
   [JP Simard](https://github.com/jpsim)
 
+* Pick the right version of declarations with type attributes.  
+  [John Fairhurst](https://github.com/johnfairh)
+  [#1148](https://github.com/realm/jazzy/issues/1148)
+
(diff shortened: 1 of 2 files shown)


jpsim:
Thanks! ❤️

Issue #829, opened 2017-05-24
Object properties without an explicit reference specifier are declared as assign


steipete:
Properties that do not have a ‘strong’ or ‘copy’ on them in source are exported as ‘assign’ in the documentation. We omit these modifiers as they are inferred automatically and the property syntax is already long enough with nullability added.

For example, this is declared in code:

```
/// Stamp image. Defaults to nil. Set to render an image.
/// @note An image set will take priority over the internal appearance stream.
@property (nonatomic, nullable) UIImage *image;
```

This is what we get:

```
@property (assign, readwrite, nonatomic, nullable) UIImage *image;
```

https://pspdfkit.com/api/ios/Classes/PSPDFStampAnnotation.html#/c:objc(cs)PSPDFStampAnnotation(py)image

It's been on our back burner for a while to fix, but haven't found time so far. Do you have a starting point for me how this can e fixed?

johnfairh:
The extra attributes are introduced by sourcekitten [here](https://github.com/jpsim/SourceKitten/blob/c01833de782d2e8866f3ada33ed266aa585c691a/Source/SourceKittenFramework/Clang%2BSourceKitten.swift#L87) -- they show up only if the decl has a doc comment which gives the code access to the fuller version of the declaration that the toolchain knows, rather than what has been parsed from the source file.

The Swift version stores both what the user typed (`parsed_declaration`) and what the tools know (`doc.declaration`) and jazzy prefers the former for display.  Maybe SK should do the same for ObjC?

jpsim:
Yeah, we get those from libclang. Maybe there's an extra flag we can pass it to help it better assume what default property attributes to add?

johnfairh:
The libclang in Xcode 11.4 makes this worse by also adding `unsafe_retained` [(commit)](https://reviews.llvm.org/D57965), so the line above comes out
```
@property (nonatomic, assign, unsafe_unretained, readwrite, nullable) UIImage *image;
```

So, tentatively propose to strip out the default property attributes `atomic`, `readwrite`, `assign`, `unsafe_retained` -- this *seems* to be how apple docs work (or they could have a stricter in-house style about not writing them.)

Could do with an opinion from genuine ObjC developers: is this OK or is there a legitimate reason to preserve them if written/show them?  I guess would have a jazzy option to disable property attribute stripping.

[...]

PR #1158, opened 2020-02-07
Xcode 11.4/Swift 5.2 fixes

johnfairh:
~Incomplete~ Workarounds and fixups for Xcode 11.4

Swift changes:
* SourceKit used to say “internal” when it didn’t know the ACL.  Now it leaves the field blank which is technically more honest but practically less helpful.  Workaround in jazzy to restore previous behaviour.
* ‘mutating’ keyword is duplicated. [SR-12139](https://bugs.swift.org/browse/SR-12139).
* Nested type names are minimally formed -- any prefix implied by context is removed.
* More properties have { get } suffixes -- getting close to being able to drop the ACL attrs
* Multiple properties defined on the same line of code with a shared attribute “public var a, b” are reported properly!
* Generic constraints more aggressively minimized -- if you write an essay that evaluates to “== Int” then that’s how it reads.
* ~`key.modulename` is broken for local dependencies.  This is what’s breaking Moya.  Will try and fix Swift, https://github.com/apple/swift/pull/29702, but may have to work around in sourcekitten.~ Fixed in beta 3.

ObjC changes:
* Many properties are now showing as `unsafe_unretained`. 
    * Changed jazzy to strip default attrs per #829.
* The order in which libclang prints out property attributes has changed. /sob

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -17,7 +17,7 @@ jobs:
 
   swift:
     macos:
-      xcode: "11.2.0"
+      xcode: "11.4.0"
     environment: *bundler-environment
     steps:
       - checkout
@@ -29,7 +29,7 @@ jobs:
 
   objc:
     macos:
-      xcode: "11.2.0"
+      xcode: "11.4.0"
     environment: *bundler-environment
     steps:
       - checkout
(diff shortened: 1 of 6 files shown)


johnfairh:
Turns out the objc `unsafe_unretained` is intentional, from [this commit](https://reviews.llvm.org/D57965).  Maybe revisit the 'strip out default property attributes' issue.

Issue #1156, opened 2020-01-30
LaTeX in markdown


arguiot:
As Apple works on SwiftNumerics and Google on Tensorflow for Swift, many math libraries will start to appear.

In order to make life easier for developers and newcomers alike, support for LaTex equations would be welcome.

I propose to use the following syntax (as for StackOverflow or others): `$<insert equation>$` for inline equations and `$$<insert equation>$$` for centered equations.

I think it won't be a challenge to implement since javascript libraries like KaTex and MathJax do that very well.

I'm submitting this issue before working on a PR to discuss about the implementation (so I don't waste my time and your time).

johnfairh:
I think it's worth exploring; Xcode quick help doesn't support it so the main use might be in guides rather than API documentation?

Were you thinking render at client or 'server' (jazzy runtime)?  Might be easier to do client-side rather than try and express the tooling dependency.  Would need an optout flag to avoid adding more JS dependencies.

Brief thought on syntax: I would probably go for hooking the backticks/code generation so you'd write markdown `` `$equation$` ``.  Implement `codespan` in the jazzy markdown renderer, spot the dollars and generate some special html tags/attrs to pick up from js at runtime.

arguiot:
I was thinking of client side rendering (way easier). Rather than having an optout flag, I think it's worth considering an opt-in flag (so it's disabled by default).
Here is an example snippet that we could use if use KaTeX (lighter, but less advanced features than MathJax):
```js
const regex = /^\`\$.*\$\`$/gm; // Check for your syntax
document.querySelectorAll('code').forEach((block) => {
                const content = block.innerHTML
               if (regex.test(content)) {
                            const eq = content.slice(2, -1)
                            katex.render(eq, block)
               }
});
```

johnfairh:
👍 for client-side.

The jazzy change I'm suggesting gives control over generating the `code` element - so instead of `<code>$equation$</code>` it could give, for example, `<span class="math">equation</span>`.  Then the selector is easier and the styling will be easier because there's no `<code>` element to fight with (unless KaTeX deals with that).

arguiot:
As I'm not a Ruby developer, I have no idea how to do that. For me it seems easier to do all with KaTeX (as I know JavaScript well). I'll probably create a PR later this week, and you'll be able to help me with that

arguiot:
As the new PR is likely to be merged very soon, I close this issue

PR #1159, opened 2020-02-14
Rendering LaTeX in markdown

arguiot:
As discussed in #1156 , LaTeX rendering inside a code block like:
```
$$
Equation
$$
```
Or `$Equation$` would be an interesting feature to add, as Apple works on SwiftNumerics and Google on Tensorflow for Swift, many math libraries will start to appear.

This PR focuses on front end rendering, @johnfairh suggested to  give control over generating the `code` element - so instead of `<code>$equation$</code>` it could give, for example, `<span class="math">equation</span>`. Then the selector is easier and the styling will be easier because there's no `<code>` element to fight with.


arguiot:
Also, I'm not sure, but maybe we also should consider passing `displayMode: true` to KaTeX, since this will center the math on the page on its own line.

arguiot:
The build is failing because it's comparing the output to a prebuilt version, so if you guys could update that so it succeed

lib/jazzy/themes/apple/assets/js/jazzy.js
```diff
@@ -57,3 +57,17 @@ $("a:not('.token')").on('click', function() {
...
+document.querySelectorAll('code').forEach((block) => {
+                const content = block.innerHTML
+               if (smallBlockRegex.test(content)) {
+                            const eq = content.slice(1, -1)
+                            katex.render(eq, block)
+               } else if (bigBlockRegex.test(content)) {
+                            const eq = content.slice(2, -2)
+                            katex.render(eq, block)
+               }
+});
```
johnfairh: Regexps: `g` flag is not required and will confuse with its `lastIndex` behaviour.  Look at the `s` flag for the `\n|.`pattern?  Missing `^$` anchors in the `bigBlockRegex` pattern?

Needs to not match in `<pre><code>` elements -- that is supposed to be verbatim.

Should test the `bigBlockRegex` first otherwise small will match (the accidental `g` flag sometimes saves it!)

Might be better to write a single regexp and use a capture group to pull out the equation to render: what's here gets confused by trailing newlines.

If katex fails then it would be good to fall back to what the user wrote - when I tried I just got blank content.

Style: looks to me as though this file uses semicolons, 2-space indent, and jQuery?

johnfairh:
Ok that came out easily enough -- https://gist.github.com/johnfairh/a82dc7cfc927491f31a3d213f8a4a52b
Feel free to tweak the styles etc. / critique my regexes :-)

arguiot:
@johnfairh the regex looks good to me, I will test with my project and tell you if it works

arguiot:
@johnfairh I'm having trouble running `bundle exec rake rebuild_integration_fixtures`.

So, I runned `bundle exec rake swift_spec` (and added a random function in MiscJazzyFeatures), but your ruby function `codespan` doesn't seem to have an effect on the final build.

I fixed a few things on the project:
- revert the ImgBot commit
- Copied a local version of KaTeX
- Fixed jazzy.js so it's compatible with your ruby function

Gemfile.lock
```diff
@@ -191,4 +191,4 @@ DEPENDENCIES
   webmock
 
 BUNDLED WITH
-   1.17.3
```
arguiot: Oops, I might broke the build with that!

johnfairh:
Hi @arguiot, the Ruby code is removing the $, and CSS class names need a dot in selectors.  So the js should be something like:

```javascript
// KaTeX rendering
document.querySelectorAll('.math').forEach((block) => {
  const content = block.innerHTML
  katex.render(content, block)
});
```
...which works for me on your fork, html:
```html
<p>How about <span class='math m-inline'>\Gamma(n) = (n-1)!\quad\forall n\in\mathbb N</span> ?</p>
```
Results:
Screenshot 2020-03-02 at 10 50 36

Yes, undo the Gemfile.lock change.

`bundle exec rake swift_spec` is a different command: it does not update the test results, it produces output in the jazzy/tmp directory -- I guess your changed files will be in there?

What is the trouble you are having with `bundle exec rake rebuild_integration_fixtures` ?

If you can add the tests, fix the styling (center/displaymode, anything else?), tweak error handling if possible, then I'll look after the "add a flag to avoid bloating every jazzy site with katex" part.

arguiot:
@johnfairh Alright, I fixed the problem with `bundle exec rake rebuild_integration_fixtures` (it was a problem with my local version of ruby in fact) and submitted a new PR https://github.com/arguiot/jazzy-integration-specs . I also fixed the styling.

johnfairh:
@arguiot -- just to let you know I'm not ignoring this, slowly working through the pieces.

arguiot:
No problem

johnfairh:
Continued in #1164.

PR #1164, opened 2020-03-17
Use KaTeX to render math expressions in markdown

johnfairh:
#1159 continued.
- Bundle KaTeX only if the generated docs require it.  No need for a flag.
- Use KaTeX options to improve display style, means no need for custom css.
- Add some docs.

Screenshot 2020-03-16 at 22 00 48

@arguiot please take a look.


arguiot:
Alright, it looks good to me. Do I have to close my PR for the integration specs?

johnfairh:
Let's merge this while Xcode 11.4 dawdles somewhere.  Not sure what is up with the CircleCI GitHub hooks, it's all green on the other end.

@arguiot yes I couldn't really use that PR because I refactored how the files were stored in the themes.  Thanks for your work on this.

PR #1167, opened 2020-03-30
Allow #//apple_ref/… links to expand docs too

pcantrell:
Currently Jazzy generates `<a name=…>` tags for both USR-style anchors and Dash-style `apple_ref` anchors:

```html
<a name="/s:6Siesta8ResourceC4loadAA7Request_pyF"></a>
<a name="//apple_ref/swift/Method/load()" class="dashAnchor"></a>
<a class="token token-open" href="#/s:6Siesta8ResourceC4loadAA7Request_pyF">load()</a>
```

The `apple_ref` anchors have several advantages: they are human-readable, generally make for shorter links, and (most importantly) are less prone to changes between doc generation cycles and thus make for more stable long-term links from e.g. other documentation.

However, they have a shortcoming in Jazzy: a link of the form `Foo.html#//apple_ref/swift/Method/load()` currently scrolls to the correct item, but does not expand it.

This PR also expands the item when thus linked, just as `Foo.html#/s:6Siesta8ResourceC4loadAA7Request_pyF` would.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -23,6 +23,11 @@
   [John Fairhurst](https://github.com/johnfairh)
   [#995](https://github.com/realm/jazzy/issues/995)
 
+* Support Dash-style `apple_ref` links to specific API items, for more
+  stable and human-readable links from external docs.
+  [Paul Cantrell](https://github.com/pcantrell)
+  [#1167](https://github.com/realm/jazzy/pull/1167)
+
 ##### Bug Fixes
 
 * Don't generate documentation if the `xcodebuild` command fails.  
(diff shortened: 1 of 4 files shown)


pcantrell:
Also, Jazzy should consider symlinking the js files that are identical across the 3 themes instead of copying them.

johnfairh:
Thanks for this -- when I try it though I can see that the `apple_ref` links work nicely but the USR ones don't any more, maybe because the `next('.token')` business doesn't go anywhere if the `a[...` element actually is the `.token`?

OK from me to merge this with a fix for that and a changelog update.  (I don't know why circle hasn't rebuilt this after your specs update...)

pcantrell:
Ah, indeed, it turns out that I was looking for `nextAll()` instead of `next()`. Good catch!

It’s also apparently possible to do this all in one selector:

    $(`a[name="${location.hash.substring(1)}”] ~ .token`)

…but that seems to me too clever by half.

johnfairh:
Great, thanks.

PR #1173, opened 2020-04-11
Fix module version not being used from podspec

Jon889:
version_configured was being checked [here](https://github.com/realm/jazzy/blob/master/lib/jazzy/doc.rb#L46) to see if the version was set but if the --podspec option was used then this check would be false so the version wasn't used.

I'm not sure if the checks in doc.rb should be changed to check if the `config.version` is empty instead?

Also I'm not sure if the other `_configured` variables such as `author_name_configured` should be set to true as well for consistency or if that might break something?

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,7 +10,8 @@
 
 ##### Bug Fixes
 
-* None.
+* Fix module version not being used from podspec.  
+  [Jonathan Bailey](https://github.com/jon889)
 
 ## 0.13.2
 
--- a/lib/jazzy/podspec_documenter.rb
+++ b/lib/jazzy/podspec_documenter.rb
@@ -59,6 +59,7 @@ module Jazzy
       end
       unless config.version_configured
         config.version = podspec.version.to_s
+        config.version_configured = true
       end
       unless config.github_file_prefix_configured
         config.github_file_prefix = github_file_prefix(podspec)
--- a/lib/jazzy/podspec_documenter.rb
+++ b/lib/jazzy/podspec_documenter.rb
@@ -45,6 +45,7 @@ module Jazzy
 
     # rubocop:disable Metrics/CyclomaticComplexity
     # rubocop:disable Metrics/PerceivedComplexity
+    # rubocop:disable Metrics/MethodLength
     def self.apply_config_defaults(podspec, config)
       return unless podspec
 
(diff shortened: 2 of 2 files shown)


Jon889:
I'm not entirely sure how to deal with the tests failing. They are failing because now the version number is appearing in the documentation, but I think this should be the correct behaviour. I'm not sure how to update the tests so that the version number is expected?

johnfairh:
Wow this is a bit of a mess.  It looks to me like always setting 'configured' when `podspec_documenter` sets things is correct: semantically we're intentionally overriding the default.  There aren't that many uses of the '_configured' accessor and they all look safe to me.

`config.version` has a  historical docset-related default of 1.0 so it is never empty.

To update the integration spec files you need to:
1. fork https://github.com/realm/jazzy-integration-specs
2. check out a new branch in jazzy/spec/integration_specs submodule
3. do `bundle exec rake rebuild_integration_fixtures` to make the updates
4. push your specs branch to your specs fork and make a PR
5. add a commit to _this_ PR with the submodule update

More detailed instructions [here](https://github.com/realm/jazzy/blob/master/CONTRIBUTING.md#updating-the-integration-specs) but use your own fork instead of the master.

Let me know if you get stuck and I'll sort it out.

Jon889:
Thanks, I’ll set the other _configured to true then and update the specs

Jon889:
I've managed to do that but I'm getting a bunch of unrelated changes as well. So I tried running it with this repo checked out on master and it still had those unrelated changes. https://github.com/realm/jazzy-integration-specs/compare/master...Jon889:jon889-nochanges?expand=1

johnfairh:
Could you have Xcode 11.3 (or something else with Swift 5.1) selected instead of 11.4?

Jon889:
Ah yeah, my work still requires Xcode 11.3, I even read that in the instructions you linked to and thought I must remember to switch 🤦

Jon889:
I've put the PR up https://github.com/realm/jazzy-integration-specs/pull/110 I've scrolled through it slowly and made sure it's only the version number changed (99% sure nothing else has :))

johnfairh:
Looks good - should be able to get CI to pass here by adding a commit with the specs/integration_specs update.

Jon889:
Ah ok, I guess I should update this PR again with the commit when the specs are merged to master?

johnfairh:
Nope, that should be it as long I do the right type of merge.  Thanks for doing all this!

Jon889:
Thanks for accepting it and helping me with it!

PR #1179, opened 2020-04-20
Force pages for specific types

bamx23:
It's my first contribution here so I hope that it is useful.
The purpose of this change is to allow page generation for specific types, like constants or type definitions. 

For example, I want to use this regex of type kinds:
```regex
^(sourcekitten\.)?source\.lang\.(objc|swift)\.decl\.(class|enum|function.*|protocol|struct|var\..*|category|constant|typedef)$
```

I haven't found any issue related to this problem. Hope I haven't forgotten to update anything. If there should be some tests point me where I should add them.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,11 @@
   typeahead.js 1.3.1 (`fullwidth` theme only).  
   [John Fairhurst](https://github.com/johnfairh)
 
+* Added a config option to specify type kinds that should be rendered 
+  as pages. The option is `--render-as-page` and it accepts a regular 
+  expression matching type kinds.  
+  [Nikolay Volosatov](https://github.com/bamx23)
+
 ##### Bug Fixes
 
 * None.
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,9 +10,9 @@
   typeahead.js 1.3.1 (`fullwidth` theme only).  
   [John Fairhurst](https://github.com/johnfairh)
 
-* Added a config option to specify type kinds that should be rendered 
-  as pages. The option is `--render-as-page` and it accepts a regular 
-  expression matching type kinds.  
+* Added a config option `--[no-]extra-page-rendering` to enforce page
+  rendering for definitions meaned global (like classes, structs, enums etc.)
+  even if they don't have children.  
   [Nikolay Volosatov](https://github.com/bamx23)
 
 ##### Bug Fixes
--- a/lib/jazzy/config.rb
+++ b/lib/jazzy/config.rb
@@ -436,6 +436,18 @@ module Jazzy
                    'is "Undocumented", put "" if no text is required',
       default: 'Undocumented'
 
+    config_attr :render_as_page_kinds,
+      command_line: '--render-as-page REGEX',
+      description: 'Regular expression that match type kinds for which page generation should be forced',
+      default: '',
+      parse: ->(r) { 
+        if r.to_s.empty?
+          nil
+        else
+          Regexp.new(r) 
+        end
+      }
+
     # rubocop:enable Style/AlignParameter
 
     def initialize
--- a/lib/jazzy/config.rb
+++ b/lib/jazzy/config.rb
@@ -436,18 +436,15 @@ module Jazzy
                    'is "Undocumented", put "" if no text is required',
       default: 'Undocumented'
 
-    config_attr :render_as_page_kinds,
-      command_line: '--render-as-page REGEX',
-      description: 'Regular expression that match type kinds for which page '\
-                   'generation should be forced',
-      default: '',
-      parse: ->(r) {
-        if r.to_s.empty?
-          nil
-        else
-          Regexp.new(r)
-        end
-      }
+    config_attr :page_render_level,
+      command_line: '--page-render-level [no-children | global]',
+      description: 'When definition should be rendered as a page. '\
+                   'Value "no-children" means that only definitions with '\
+                   'no children should have their pages. Value "global" '\
+                   'means that pages will be generated for both '\
+                   'no-children-definitions and definitions meaned to be '\
+                   'global like classes, structs, enums etc.',
+      default: 'no-children'
 
     # rubocop:enable Style/AlignParameter
 
--- a/lib/jazzy/config.rb
+++ b/lib/jazzy/config.rb
@@ -436,15 +436,12 @@ module Jazzy
                    'is "Undocumented", put "" if no text is required',
       default: 'Undocumented'
 
-    config_attr :page_render_level,
-      command_line: '--page-render-level [no-children | global]',
-      description: 'When definition should be rendered as a page. '\
-                   'Value "no-children" means that only definitions with '\
-                   'no children should have their pages. Value "global" '\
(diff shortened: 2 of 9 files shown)


johnfairh:
Thanks for the PR.

You can run rubocop (a linter) locally by doing 'rake rubocop' to see what needs to be done to fix CI.  See CONTRIBUTING.md for the changelog format - two spaces (markdown newline) needed at the end of the description.

You can add tests to the `misc_jazzy_features` project, see #1173 for recent notes on how to do that.

The functional problem with the PR so far is filename uniqueness.  Jazzy today relies on the Swift/C namespace rules for this, deriving the page filename directly and independently from the symbol name -- with this change, things like `func f(a: Int)` and `func f(a: String)` get aliased onto `f(a:).html` because they share a symbol name.

The design problem is that the parent pages look wrong:
Screenshot 2020-04-20 at 18 19 20
... the info in the disclosure section is exactly duplicating the decl's own page so the 'See more ...' link doesn't make literal sense and the pure repetition of the content feels bad/confusing.  If you look at the Apple docs that your design is trying to mirror, you can see there that they have a mini-summary on their category pages and a link to the child, nothing else.  You could try to reuse the `omit_parent_from_content` type style to get a straight link, might be OK?

The regex approach is quite undiscoverable and hard for normal users to access.  It might be better to decide on what design is good and just provide a flag to do that.  Can you explain more about your example regex?  What are you trying to exclude?  Just extensions?  How come? Why not say 'render everything on its own page'?

bamx23:
Thanks for your response and explanations.

I haven't thought about methods/functions collisions here. The main purpose of this PR is, as you correctly guessed, to "render everything on its own page". Except for methods/properties etc. I mean, everything that is global(constants, enums(with other typedefs), protocols, classes, structs, global functions etc) should have their own page even if they don't have any children. The regex that I used as an example does exactly this, as far as I may concern.

But I agree, that regex as an option is not that good for end-users, because they have to know all the types. What do you think about an option similar to `--min-acl` in the way that it will define how much pages should be generated: 
* only for types with children (current logic)
* for everything that is global (the logic I expect)
* for every type (don't think that we need this, you described why, but as an example of the third option).
The option could be named like `--page-generation-level` with possible values `with-children` and `global`.

And I didn't get the idea about `omit_parent_from_content`. If it matters after my comments above, could you explain what do you mean?

I will add all the tests later when we come to the best solution.

johnfairh:
> Except for methods/properties

Ok, I understand.  The regex matches `source.lang.swift.decl.function.method.instance` etc. which is why I thought 'everything' was the goal.  Sounds like just a typo.

Should a nested struct with no children get its own page?  (`struct A { struct B {} }`)

Having a policy/style option sounds like a good idea.

> omit_parent_from_content

If we create a page for a childless declaration like a top-level function, then I don't think we should duplicate all of that content inline in the parent page: it's confusing to have the exact same info in two places.  The method `omit_parent_from_content` does that - if it's `true` then you just get a link from the parent into the child page, without the popopen box.  Jazzy uses it only for guide index pages today, I think.

bamx23:
> Sounds like just a typo.

Yes, it was a typo. And this typo also shows the problem with regexes.

> Should a nested struct with no children get its own page?

That's a good question. I guess nested structures shouldn't have their own pages. I will experiment with this in tests.

> `omit_parent_from_content`

Oh, now I see. I suppose you mean `omit_content_from_parent`. I will experiment with this one too. Now it is always `false`. But with this new page rendering logic, this method might be useful to make everything works and looks right.

Thank you, I will experiment with all this logic and tests and come back with a better solution later.

johnfairh:
> I suppose you mean omit_content_from_parent

oops, yes exactly!

bamx23:
I've decided to simplify the new option as much as possible. I don't know if there will be "new styles" for this kind of rendering logic, so I've made a simple bool option that enables "extra rendering mode".

Tests are also made here: https://github.com/realm/jazzy-integration-specs/pull/114  
I've decided to enable this new option in both `misc_jazy_features` and `misc_jazzy_objc_features` tests. All other suits are left untouched. Is this what you expected me to do with tests?

> Should a nested struct with no children get its own page?

I've been thinking about this for a while and have come to the conclusion that empty nested classes should be rendered the same way as non-empty ones. I don't see any reason why non-nested empty classes should have their pages but nested ones don't.

johnfairh:
I noticed a couple more design things looking at the specs: function parameters & returns aren't included when the function gets its own page, and the 'see on github' link isn't included there either.

Do you think it would be consistent to include extensions in the list?  They're not always caught by the "has children" rule -- eg. `extension String : MyProtocol {}`.

bamx23:
> Do you think it would be consistent to include extensions in the list?

Yes, I think so. Added.
Nevertheless, I'm not that confident about decisions around Swift definitions. My case is mostly about ObjC.

bamx23:
@johnfairh Is there anything I should do here?

johnfairh:
I think the list of things remaining is:
1. Filename collisions on the new leaf pages;
2. Missing content from the new leaf pages;
3. Leaf page content repeated in their parent.

bamx23:
Thanks, I'll come back with answers/solutions for these ones later next week.

bamx23:
> 1. Filename collisions on the new leaf pages

I've fixed filename collisions issue (and added some example/tests in `misc_jazzy_features`). Please check if this solution is OK. I can't imagine any other collision except for global functions with different argument/return types. At least for the definitions, this PR allows generating pages for.

> 2. Missing content from the new leaf pages;

I'm not quite sure about this point. Empty classes/protocols/extensions may have reasons to exist and in this "extended rendering mode" will have small but their own pages. If you know when it can be a problem, please give me an example.

> 3. Leaf page content repeated in their parent.

This was before too. I.e. for nested classes with members. The only thing I've added is an option to provide "See more" for those definitions. Could you give an example when it can look bad?

johnfairh:
Paging some of this back in:
1. We should probably keep the urls involving the symbol name.  I think you're right, at the moment it's only functions that can collide
2. I guess this is about first para of [this comment](https://github.com/realm/jazzy/pull/1179#issuecomment-620220297)
3. See last para of [this comment](https://github.com/realm/jazzy/pull/1179#issuecomment-617385425), seems like that's the source of the requirement

bamx23:
Thanks for the clarification, the last two points are absolutely reasonable. I've checked how we can include all information from non-leaf pages inside childless pages and realized that current templates are just not built for that. The `doc` template does not include this information, it is stored in the `item` template. But it doesn't look right if I include `item` template inside `doc`, because this creates a foldable block. So I've just moved all missing parts from `item` template into `doc`. I've tried to make some generalization with a new reusable template, but it seems to be much more 'breaking' change.

Talking about the first point about file names. I'm not sure we can keep previous names. But because with the new flag jazzy will generate new pages it doesn't break any compatibility.

johnfairh:
Thanks for keeping on with this, must have blown up beyond your original thought.  I think what you've done with the templates is fine, agree refactoring the top-level/nested items is best left out of this one.  I'll find some time this week to merge this all up.

bamx23:
Thanks for supporting and explaining all the pitfalls. Looking forward to the upcoming merge.

johnfairh:
Continued in #1200.

PR #1200, opened 2020-06-05
Add flag to always create pages for classes etc.

johnfairh:
Tidy of #1179 :
* Rename flag
* Various wordsmithing
* Various Ruby/DRY refactors
* Adjust CSS for ugly vertical spacing of GitHub links
* Fix theme HTML for inline links to nested or deprecated types

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -207,7 +207,11 @@
 
 ##### Enhancements
 
-* None.
+* Added option `--[no-]separate-global-declarations` to always create separate
+  documentation pages for top-level declarations as well as classes,
+  structures, enums etc. even if they don't have members.  
+  [Nikolay Volosatov](https://github.com/bamx23)
+  [John Fairhurst](https://github.com/johnfairh)
 
 ##### Bug Fixes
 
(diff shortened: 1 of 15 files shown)


johnfairh:
@bamx23 could you take a look at this version please & check I haven't broken anything?

bamx23:
@johnfairh Thanks, this code looks much more clear. I have no production experience in Ruby, so I really appreciate you've refactored it. 

The code works as expected. But there is one thing. I've checked it against my ObjC library and have found that we don't show Swift declaration as we do for declarations that don't have their own pages. What do you think? Maybe it's possible to place both Swift and ObjC declarations into new separated pages?

johnfairh:
> place both Swift and ObjC declarations into new separated pages?

Yes I see what you mean (I think) -- looks like another piece of `item` missing from the top-level version.  I've added it with a tweak to keep the current behaviour of not showing the language name if there is only one language.



The specs changes are terrifying because of a new blank line - `git diff --ignore-blank-lines --ignore-space-at-eol` to see the actual changes.

bamx23:
Thanks, I like how it looks now. 
I have no more questions about this PR. Looking forward to see it merged.

johnfairh:
Right - let's merge this before it grows any more heads.

bamx23:
@johnfairh when a new release with these changes is supposed to be made?

johnfairh:
Probably after WWDC

PR #1212, opened 2020-07-30
Improve Swift API docs for mixed language projects

steviki:
This PR hides API that is marked with `NS_REFINED_FOR_SWIFT` from `display_other_language_declaration`, and therefore in API documentation.

SourceKitten doesn't seem to expose if an API is refined for Swift, so I added a check for double underscores in the `display_other_language_declaration` getter.

### DIFF
--- a/lib/jazzy/source_declaration.rb
+++ b/lib/jazzy/source_declaration.rb
@@ -105,9 +105,11 @@ module Jazzy
     def display_other_language_declaration
       # Don't expose if API is NS_REFINED_FOR_SWIFT
       is_refined_for_swift = swift_name.to_s.include?('__')
-      
+
       return other_language_declaration unless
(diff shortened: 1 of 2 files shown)


steviki:
I updated the integration specs at https://github.com/realm/jazzy-integration-specs/pull/120, but I didn't bump the submodule in this branch, since I'm not sure what the procedure here is exactly.

johnfairh:
I'm not sure this is OK because identifiers can legitimately contain underscores.

However!  Xcode 12 / Swift 5.3 has a change to SourceKit / the Clang importer that prevents `NS_REFINED_FOR_SWIFT` declarations from being seen at our level, which will automatically do what this PR wants without false positives.

steviki:
@johnfairh You are right. This might indeed lead to false positives.

That's great to hear! I guess we can wait a bit longer to get this automatically once Xcode 12 is used. Thanks for your reply.
I will close this PR then.

Issue #1246, opened 2021-04-23
Kramdown footnotes don't seem to work?


aufflick:
Jazzy uses kramdown (according to the git logs that I've skimmed), and Kramdown has a footnote syntax: https://kramdown.gettalong.org/quickref.html#footnotes But when I try to use that syntax in raw .md files included by Jazzy, the HTML just includes literal "[^1]" body text.

I don't see any options that Kramdown needs to render this.

johnfairh:
Jazzy uses [redcarpet](https://github.com/vmg/redcarpet) for its markdown rendering.  It looks like there's a footnote option available that isn't turned on -- I haven't checked how the default HTML renderer copes with it, would be worth investigating.

aufflick:
Ah I see - I didn't notice there were two Markdown parser libraries in use :) I'll play with enabling that option and see if there are any negative consequences.

johnfairh:
I had a go at this in #1253.

aufflick:
This is amazing - I just tried it out and it works perfectly :D

johnfairh:
Great - will be in next release.

PR #1253, opened 2021-05-05
Support markdown footnotes

johnfairh:
From #1246.  Redcarpet works OK when the flag is turned on but need various adjustments for themes and to allow multiple markdown blocks to be put on the same web page.

Add example to specs, all specs update because of CSS changes.


aufflick:
FWIW I tested locally with my content and it works perfectly.

PR #1254, opened 2021-05-06
Release 0.13.7 and update gems

johnfairh:


johnfairh:
Well Danger is definitely working again... Should teach it about version changes.

PR #1257, opened 2021-06-15
Simple DocC markdown syntax support

johnfairh:
DocC adds some features to markdown; it's worth supporting these in Jazzy to help users migrate their projects and get used to writing things in the new way as we all transition (however many years that takes!)

- Links.  DocC symbol links have various [features](https://developer.apple.com/documentation/xcode/formatting-your-documentation-content#Link-to-Symbols-and-Other-Content). We can recognize the syntax and honor them as autolinks most of the time.  This means users can write declaration links that work in quick help, docc, and jazzy.
- Callouts.  DocC hijacks blockquote syntax for [another form of callout](https://developer.apple.com/documentation/xcode/formatting-your-documentation-content#Add-Notes-and-Other-Asides).  It's easy enough to support this while retaining traditional blockquotes

Main thing not addressed is the addition of simple description lists; we could support these but it's a pain and the unadulterated rendering is fairly legible:

* term term1: desc1
* term term2: desc2

Spec updates for tests, some html whitespace.  Moya presciently using the new callout-blockquote -- looks fine.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,7 +6,8 @@
 
 ##### Enhancements
 
-* None.
+* Support DocC-style autolinks and callouts in markdown.  
+  [John Fairhurst](https://github.com/johnfairh)
 
 ##### Bug Fixes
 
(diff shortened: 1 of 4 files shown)


jpsim (approved):
DocC 🤝 Jazzy

PR #1267, opened 2021-07-19
Adding a license header to Jazzy JS files

hyandell:
Hi Jazzy folk,

I noticed that while the other open source files Jazzy copies into the output identify their licensing, the ones from Jazzy themselves don't. I've taken the liberty of suggesting a source header to add to each of them.


johnfairh:
Thanks for the prompt.  Doing this via #1269 to include css and CI.

PR #1278, opened 2021-09-16
Run CI on Xcode 13 / Swift 5.5

johnfairh:
Something odd going on with line numbers coming out of SourceKit but the result is better 🤷🏻 .

Realm ObjC changes because the importer is generating `async` Swift functions for me -- I have macOS 12 locally so expect (hope) CI on macOS 11 will fail.  GitHub Actions doesn't seem to know anything about macOS 12....

Something odd going on with `xcodebuild` meaning I need to explicitly set a destination to avoid a spammy warning.

Swift 5.5 has changed how `typealias`s are treated in interface generation to "look through" the typealias more often, meaning the declaration comes out differently to what the user has written.  Affects Realm Swift.  Will pursue upstream; affects DocC etc. too.

### DIFF
--- a/.github/workflows/Tests.yml
+++ b/.github/workflows/Tests.yml
@@ -31,7 +31,7 @@ jobs:
       matrix:
         spec: ["objc_spec", "swift_spec", "cocoapods_spec"]
     env:
-      DEVELOPER_DIR: /Applications/Xcode_13.app/Contents/Developer
+      DEVELOPER_DIR: /Applications/Xcode_13.0.app/Contents/Developer
     steps:
       - uses: actions/checkout@v2
         with:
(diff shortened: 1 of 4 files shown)


johnfairh:
Baffled by this:
Screenshot 2021-09-16 at 18 36 26
CI is on macOS 11, Xcode 13.0 has only a macOS 11.3 SDK .. so I suppose somehow it's defaulting min-deployment-target of macOS 12?

[SR-15202](https://bugs.swift.org/browse/SR-15202) for the typealias printing bug - I'm pretty sure it is a bug, it only happens in generic constraints.

johnfairh:
For my reference: generating the module interface for ObjC like this generates both Swift APIs, one with `async` and one traditional.  When we xref the ObjC's USR though we get the `async` one back -- the find-usr API only returns one value.  Need to check whether these two decls really share the same USR, if there's anything sourcekitten can do to retrieve them both...

PR #1294, opened 2021-11-29
Support pre-generated Swift symbolgraph files

esteluk:
Since Swift 5.5, the Swift compiler has supported the `-emit-symbol-graph`
and `-emit-symbol-graph-dir` flags to directly create Symbol graph files
as part of the compilation process. If a developer has access to these
files, it's not necessary for Jazzy to use `symbolgraph-extract` to attempt
to separately generate them.

This commit adds a `--symbolgraph_directory` configuration option to Jazzy.
If this option is provided, Jazzy skips the generation of symbolgraph files
and directly parses any `*.symbol.json` files in the given directory.

- [x] Check if any updates need to be made to test specs


esteluk:
Sorry, meant to raise this initially on my fork rather than here, but having done so it might as well stay open...

I found this a valuable change because I was struggling a little bit to find the correct `build-tool-arguments` that would generate valid symbol files for my framework, particularly when using the Simulator SDK. Once I realised that the build process had already spit out the required files, it seemed like it would be useful for Jazzy to be able to consume them directly. 

If this is a direction that you'd prefer not to go down, I'd be happy to close the PR.

CHANGELOG.md
```diff
@@ -9,7 +9,8 @@
 
 ##### Enhancements
 
-* None.
+* Support using pre-generated symbolgraph files in Swift symbolgraph mode.
```
johnfairh, line 12: Need end-of-line spaces for markdown, see `CONTRIBUTING.md`.

README.md
```diff
@@ -260,6 +260,14 @@ Some examples:
    This implies that `/Build/Products/MyMod.framework` exists and contains
    a `.swiftmodule`.  Again the `--source-directory` is searched by default
    if `-F` is not passed in.
+5. With pre-generated symbolgraph files:
+    ```shell
+    jazzy --module MyMod --swift-build-tool symbolgraph
+          --symbolgraph_directory Build/symbolgraphs
```
johnfairh, line 266: Should be `--symbolgraph-directory` here?  With dash not underscore?

README.md
```diff
@@ -260,6 +260,14 @@ Some examples:
...
+    ```
+    If you've separately generated symbolgraph files by the use of 
+    `-emit-symbol-graph`, you can pass the location of these files using
+    `--symbolgraph_directory` from where they can be parsed directly.
```
johnfairh, line 270: (same)

lib/jazzy/symbol_graph.rb
```diff
@@ -69,6 +67,20 @@ def self.arguments(config, output_path)
       args + config.build_tool_arguments
     end
 
+    # Parse the symbol files in the given directory
+    def self.parseSymbols(directory)
```
johnfairh, line 71: snake-case for Ruby, `self.parse_symbols`

spec/integration_spec.rb
```diff
@@ -238,6 +240,19 @@ def configure_cocoapods
                             '--swift-build-tool symbolgraph ' \
                               "--build-tool-arguments -I,#{module_path} "
     end
+
+    describe 'Creates docs for Swift project from symbolgraph files' do
```
johnfairh, line 244: I'd prefer to not change this file, it's quite a lot of overhead to get this working and adding a entire new output set adds change-review burden in the future as well.  I appreciate the effort to add some kind of test for the feature but jazzy doesn't have a way of doing convenient unit-testing.

johnfairh (changes requested):

johnfairh:
I'm happy to merge this option -- left a few comments.  I don't know what's up with Danger in CI, maybe we have something misconfigured for forks, but you should be able to run `rake rubocop` locally.  Don't worry if Danger doesn't sort itself out on your next commit, I'll take care of it.

README.md
esteluk: Yeah sorry, I'd been staring at the yaml files too much!

spec/integration_spec.rb
esteluk: Fair enough, that makes sense. I was hoping that it would be possible to reuse the same before/after states as the main symbolgraph tests, but I think there are some differences in output that would make this impossible. Giving it half-of another shot, but remove this test shortly otherwise.

johnfairh:
Turns out Danger + GitHub actions is broken on forks due to secret handling.

johnfairh:
All done - thanks for the PR!

PR #1298, opened 2022-02-03
Swift 5.6

johnfairh:
updated sourcekitten for sourcekit change

jazzy_swift - improved `nonisolated` decl printing

realm_objc - new `@unchecked Sendable` conformances for ObjC types

jazzy_objc - Sendable, plus an interesting consequence of the module-name changes that I’ll leave unchanged (module name in .jazzy.yaml doesn’t match SPM’s opinion) for coverage.

realm_swift - probably a Swift bug to do with `override` vs. `@objc`, opened [SR-15815](https://bugs.swift.org/browse/SR-15815)

### DIFF
--- a/.github/workflows/Tests.yml
+++ b/.github/workflows/Tests.yml
@@ -31,7 +31,7 @@ jobs:
       matrix:
         spec: ["objc_spec", "swift_spec", "cocoapods_spec"]
     env:
-      DEVELOPER_DIR: /Applications/Xcode_13.3.app/Contents/Developer
+      DEVELOPER_DIR: /Applications/Xcode_13.2.app/Contents/Developer
     steps:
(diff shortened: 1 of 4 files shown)


jpsim:
Might be worth updating the `sourcekitten` step in the Rakefile to build a universal binary:

```
swift build -c release --arch arm64 --arch x86_64
```

Thanks for doing this!

johnfairh:
Oh will do -- despite cross-compiling all the time in other contexts my brain had decided that I couldn't do this without an M1 system...

johnfairh:
Because Xcode 13.3 needs macOS 12 and GitHub Actions doesn't support it, leave CI on Swift 5.5.

So this PR becomes just updating sourcekitten to a universal binary of its latest release.

PR #1299, opened 2022-02-19
Objective C docs

johnfairh:
Braindump of Objective-C docs generation techniques and workarounds.

Add a warning for a probably-doomed combination of options.


jpsim:
Thanks so much for doing this, John 👏

PR #1331, opened 2022-10-19
Improve README.md regarding Xcode selection

chkpnt:
Using `--swift-version` can take some time. Using `DEVELOPER_DIR` is much faster.


johnfairh:
Thanks!  I'm not a huge fan of this flag but it's been there forever.

PR #1334, opened 2022-11-19
Move spec tests to run on Ruby 3

johnfairh:
Ruby has been doing a painful change to how keyword arguments are passed.

Whatever, there’s a breaking change in Ruby 3 which I want to support for dev purposes to avoid requiring increasingly unsupported versions.

CLIntegracon which is borderline unsupported itself is a victim of this change.  Cocoapods is probably the only other serious user and has got a fix into CLIntegracon trunk.  Thanks Sam G again.

This PR picks up that latest CLIntegracon in the same way as Cocoapods has done.  This ends up jumping quite a few dozen commits requiring some (maddeningly) subtle changes to our driver.

There’s one test output change caused by a weird interaction between `swift-driver` and CLIntegracon — the new version is better, it looks like previously some user-facing output mysteriously vanished.

### DIFF
--- a/.github/workflows/Tests.yml
+++ b/.github/workflows/Tests.yml
@@ -13,7 +13,7 @@ jobs:
       - uses: actions/checkout@v2
       - uses: ruby/setup-ruby@v1
         with:
-          ruby-version: 2.6
+          ruby-version: 3.0
           bundler-cache: true
       - name: Rubocop
         run: |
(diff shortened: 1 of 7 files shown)


johnfairh:
Accidentally moved CI to Ruby 3.1 which is fine but shows:
* bundler update required to work around a Ruby 3 incompatibility
* an actual reproducibility/stability bug in jazzy symgraph, need to sort a list of files (`Dir[]` guaranteed sorted on Ruby 3 but not previously...)

johnfairh:
Finally, `actions/cache@v2` is grumbling about deprecations, update all actions to latest.

PR #1345, opened 2023-03-21
Create SECURITY.md

Theholycityweb:
intentionally generic


johnfairh:
If there's something specific then you could add it to the readme.

PR #1356, opened 2023-06-25
Move CI to Swift 5.9 / Xcode 15 beta

johnfairh:
Move to Swift 5.9 so I can get tests written for 5.9 features while I have the time.

Specs changes:

**realm objc**
`libdispatch` types not being translated from their ObjC names into their Swift names.
Visible in Xcode 15b2 as well.  Something wrong with the macos SDK in Xcode 15, works fine using Swift 5.9 against a 14.3 sysroot.

**moya/siesta**
_no such module `AppKit` / `UIKit.UIImage`_ reports.
This is due to a Swift 5.9 SourceKit change — if we request cursorinfo for a symbol excluded from compilation by eg. `#if os(iOS)` then Swift attempts to compile that branch of the `#if` a bit harder than earlier, which fails.  Does not affect the actual sourcekit behaviour.

**symgraph**
Better treatment of `@spi` declarations. Things have moved around because previously these things didn’t even get locations; now we know where they should go.

### DIFF
--- a/.github/workflows/Tests.yml
+++ b/.github/workflows/Tests.yml
@@ -31,7 +31,7 @@ jobs:
       matrix:
         spec: ["objc_spec", "swift_spec", "cocoapods_spec"]
     env:
-      DEVELOPER_DIR: /Applications/Xcode_14.3.app/Contents/Developer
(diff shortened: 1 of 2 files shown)


johnfairh:
GitHub actions finally moved past Xcode15 beta 1 so can get on with this.

Issue #564, opened 2016-05-19
Combine Generated Documentation


gngrwzrd-intel:
This is probably a crazy request.

It would be nice if you could combine documentation together from different projects.

I'm working on a project where each group is building an SDK and documenting it. Unfortunately they're all in separate repositories, so even though the SDKs are related and really should have documentation all in one place, they end up having separate documentation for each SDK.

It would be awesome if there was some intermediate format so you could generate documentation for frameworkA, and then generate docs for frameworkB and combine that with frameworkA documentation.

Or maybe there's a separate switch to the jazzy command to combine documentation files where it simply updates the generated HTML files by combining them?

jpsim:
This isn't crazy at all, in fact, it's how CocoaPods support works today, by building all the support platforms & subspecs, then merging them together. You can homebrew this yourself today by running `sourcekitten doc` on all the things you want to document, then merge the JSON output somehow, then feed that to jazzy's `--sourcekitten-sourcefile` parameter.

There's no conveniences to do this in a single jazzy command today though.

gngrwzrd-intel:
@jpsim That's perfect. Thanks!

AnthonyOliveri:
For anyone wondering how to implement @jpsim's suggestion, here is a simple example of how to combine API docs for a framework and one of its Cocoapods dependencies using bash. 

```bash
dependencyDocs=$(sourcekitten doc --module-name MyDependency -- -project Pods/Pods.xcodeproj)
frameworkDocs=$(sourcekitten doc)
echo ${dependencyDocs%?}', '${frameworkDocs:1} > kitty.json
jazzy --sourcekitten-sourcefile kitty.json
```

dabrahams:
I have a project in a single repo, but it's built from multiple modules, and until this has first-class support in Jazzy I'm afraid I'm forced to use Apple's DocC tool (whose support for merged documentation is not great either, but is a start).

dabrahams:
I take it back; DocC can't merge documentation from my multiple modules. @AnthonyOliveri is there an incantation similar to [this one](#issuecomment-265289632) I can use to merge the output of multiple jazzy invocations?

Thanks!

johnfairh:
Some SPM merge-module instructions it's worth linking directly from this issue [here](https://github.com/realm/jazzy/issues/1194#issuecomment-633012820).

[...]

PR #1363, opened 2023-09-07
Custom modules categories based on json files

argjiramala-tomtom:
👋 Hey jazzy community!

I wanted to share a feature proposal for enhancing the Jazzy experience for our TomTomSDK API reference. Currently, Jazzy is used to build the API references for our navigation app, but it doesn't fully meet our needs. Here's what we have and what we need:

📚 What we currently have for API Reference documentation: We have a custom index.html page that combines all the modules into a single list. You can check it out here: [link to current documentation](https://developer.tomtom.com/assets/downloads/tomtom-sdks/ios/api-reference/0.26.1/index.html)

📝 What we need: We want to provide a better user experience when navigating between APIs from different modules. We would like to have seamless navigation, similar to what Jazzy already offers within a single module.

💡 Concept: 
Screenshot 2023-09-07 at 16 51 47
Screenshot 2023-09-07 at 16 51 59

🤔 Questions:

* Does it make sense to have this as a Jazzy feature? 
* If it makes sense, how would you suggest implementing it from an architectural perspective of the current library?

📋 What does this PR (Pull Request) contain? Here's a breakdown of the changes we made:
 _* precondition (Generate .json files using SourceKitten, following the instructions in the Jazzy repository)*_

* We used the .json file names to create custom categories per file.
* We modified the 'group_docs' (labeled as 'group_docs2' in the PR) by adding an extra parameter for dynamicCategories.
* We modified the 'group_type_categories' (labeled as 'group_type_categories2' in the PR) by creating a new group based on the module name.

_* Please note that this implementation is still a work in progress and may not cover all cases, but it illustrates what we aim to achieve.*_

We appreciate your feedback and suggestions on how to improve this feature proposal.

### DIFF
--- a/lib/jazzy/doc_builder.rb
+++ b/lib/jazzy/doc_builder.rb
@@ -154,6 +154,7 @@ module Jazzy
     def self.build_docs_for_sourcekitten_output(sourcekitten_output, options)
       (docs, stats) = SourceKitten.parse(
         sourcekitten_output,
+        options,
         options.min_acl,
         options.skip_undocumented,
         DocumentationGenerator.source_docs,
--- a/lib/jazzy/sourcekitten.rb
+++ b/lib/jazzy/sourcekitten.rb
@@ -72,20 +72,9 @@ module Jazzy
     def self.group_type_categories2(docs, type_category_prefix)
         group = type_category_prefix.map do |moduleName|
           children = docs.select { |doc| doc.modulename == moduleName }
-          make_group2(children, moduleName)
+          make_group(children, moduleName, '')
         end
         [group.compact, docs]
-    end
-    
-    def self.make_group2(group, name)
-      group.reject! { |doc| doc.name.empty? }
-      unless group.empty?
-        SourceDeclaration.new.tap do |sd|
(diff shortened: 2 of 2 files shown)


johnfairh:
Yes it's a valid feature to create documentation for multiple modules at once, so you could write `jazzy --modules ModuleA,ModuleB` and get a complete docs site.

Issue #564 is the tracker for this.

The general top-level issues that I remember are:
1. File layout on disk for multi-module cases with back-compatibility for single-module
2. Autolinking between/within modules (ie. with/without module name)
3. UI - extensions of types provided by other modules
4. UI - organise/design the left nav and page breadcrumbs.  Probably two sensible defaults (merge modules like today, split modules out as in your prototype).  Deal with custom categories/abstracts

If all you want from jazzy is the output of your prototype then I think you should just use `custom_categories` to organise the combined sourcekitten JSON site -- you could write a script to generate that from the JSON if you didn't want to maintain it.

It is disappointing that DocC hasn't addressed this use case yet but Apple did talk about working on it fairly recently - there's a good chance it will appear in the next couple of years.

johnfairh:
Continued in #1379

PR #1369, opened 2023-10-10
Improving Jazzy Documentation for Multiple Modules

argjiramala-tomtom:
This pull request introduces enhancements to the Jazzy documentation tool, specifically focusing on supporting multiple modules within Xcode projects or workspaces.

To summarize the changes made:

* New Command Line Attribute: A new command line attribute called "**modules**" has been added. This attribute allows users to specify multiple module names when generating documentation using Jazzy.
Usage example:
`jazzy --modules ModuleName1,ModuleName2,...,ModuleNameN`

* Support for Workspace-based projects: If you have multiple Xcode projects that can only be successfully built using a workspace, follow these steps:

a. Each module should have its own configuration file with "**xcodebuild_arguments**" specified

Usage example:
`jazzy --modules ModuleName1,ModuleName2,...,ModuleNameN`

Proposal PR: https://github.com/realm/jazzy/pull/1363

### DIFF
--- a/lib/jazzy/config.rb
+++ b/lib/jazzy/config.rb
@@ -263,6 +263,10 @@ module Jazzy
       description: 'Name of module being documented. (e.g. RealmSwift)',
       default: ''
 
+    config_attr :multiple_modules,
+      command_line: ['--multiple_modules MODULE_NAME1,...ModuleNameN', Array],
+      description: 'Name of modules being documented. (e.g. RealmSwift,...)'
+
     config_attr :version,
       command_line: '--module-version VERSION',
       description: 'Version string to use as part of the default docs ' \
--- a/lib/jazzy/config.rb
+++ b/lib/jazzy/config.rb
@@ -178,12 +178,9 @@ module Jazzy
         'sourcekitten.',
       default: []
 
-    config_attr :custom_modules,
-      # command_line: ['-cm', '--build-tool-arguments arg1,arg2,…argN', NArray],
-      description: 'Custom navigation categories to replace the standard ' \
-        "'Classes', 'Protocols', etc. Types not explicitly named " \
-        'in a custom category appear in generic groups at the ' \
-        'end.  Example: https://git.io/v4Bcp',
+    config_attr :modules,
+      description: 'Array of modules that are going to be documented ' \
+      'It will contain arguments:  - name, build-tool-arguments arg1,arg2,…argN, source_directory.',
       default: false
 
     alias_config_attr :xcodebuild_arguments, :build_tool_arguments,
@@ -271,10 +268,6 @@ module Jazzy
       command_line: ['-m', '--module MODULE_NAME'],
       description: 'Name of module being documented. (e.g. RealmSwift)',
       default: ''
-
-    config_attr :modules,
-      command_line: ['--modules MODULE_NAME1,...ModuleNameN', Array],
-      description: 'Name of module being documented. (e.g. RealmSwift)'
 
     config_attr :version,
       command_line: '--module-version VERSION',
@@ -586,6 +579,7 @@ module Jazzy
         self.class.all_config_attrs.group_by(&prop)
       end
 
+
       config_file.each do |key, value|
         unless attr = attrs_by_conf_key[key]
           message = "Unknown config file attribute #{key.inspect}"
@@ -596,7 +590,6 @@ module Jazzy
           warning message
           next
         end
-
         attr.first.set_if_unconfigured(self, value)
       end
 
@@ -618,8 +611,9 @@ module Jazzy
           '`framework_root` or `umbrella_header` may be ignored.'
       end
 
-      if custom_modules_configured && module_name_configured
-        raise 'Usage of config and comand line'
+      if modules_configured && module_name_configured
+        raise 'Jazzy only allows the use of a single command for generating documentation.' \
+        'Using both module configuration and modules configuration together is not supported.'
       end
     end
 
@@ -632,7 +626,6 @@ module Jazzy
         candidate = dir.join('.jazzy.yaml')
         return candidate if candidate.exist?
       end
-
       nil
     end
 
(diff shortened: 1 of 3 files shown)


johnfairh:
Thanks for the PR -- raises some interesting issues about CLI/config for multiple modules.

I don't understand the workspace problem you are trying to solve: the way it's described and coded seems very specific to your personal project/environment.  Can you explain the problem you have?

On the design side:

We should have just one parameter `--modules` that can take a list of modules, with the default single-module mode falling out as a subset of that.

I think what the piece about workspace/scheme stuff reveals is that some CLI/config parameters need to be per-module scope rather than global.  I can't think how to do that on the CLI so might have to be config-file only.  So those per-module flags might include:
* `--build-tool-arguments` (which would allow your scheme/workspace stuff I guess)
* `--source-directory` (maybe - your example seems to need this but needs investigation!)
* `--swift-build-tool` (probably not -- I think it's OK to say that if you're building multiple modules then they share this)
* `--sourcekitten-sourcefile` (eg. mixed objc+swift modules)

...need to eventually go through them all.

argjiramala-tomtom:
> Thanks for the PR -- raises some interesting issues about CLI/config for multiple modules.
> 
> I don't understand the workspace problem you are trying to solve: the way it's described and coded seems very specific to your personal project/environment. Can you explain the problem you have?
>  
> On the design side:
> 
> We should have just one parameter `--modules` that can take a list of modules, with the default single-module mode falling out as a subset of that.

> 
> I think what the piece about workspace/scheme stuff reveals is that some CLI/config parameters need to be per-module scope rather than global. I can't think how to do that on the CLI so might have to be config-file only. So those per-module flags might include:
> 
> * `--build-tool-arguments` (which would allow your scheme/workspace stuff I guess)
> * `--source-directory` (maybe - your example seems to need this but needs investigation!)
> * `--swift-build-tool` (probably not -- I think it's OK to say that if you're building multiple modules, then they share this)
> * `--sourcekitten-sourcefile` (eg. mixed objc+swift modules)
> 
> ...need to eventually go through them all.

The problem with the workspace was when the project could be built only through the workspace by specifying the scheme. The issue was solved by using the jazzy config file of the module.
I have modified that part, where now when building the project it uses the module config file instead of the global config-file.

Also the renamed the parameter from: "--multiple_modules" to: "--modules"

johnfairh:
I think there are two use cases for building multiple modules:
1. All Jazzy needs is the module name, SPM/Xcodebuild will just work.  This will be very common especially for SPM and the 'App + Extensions in one Xcode project' that users often ask about.  This can work on the command line or in the config file.  This could be a new `--modules` basically just like in the PR today.
2. Jazzy needs special build flags for each module.  This is needed for Xcode with workspaces and schemes and anything with more complicated build flags (and any Objective-C module in the future).  I think this has to be config-file only, something like this where `custom_modules` is an array of objects whose keys are about the module:

```yaml
custom_modules:
 - name: ModuleA
   build_tool_arguments: [ '-workspace', 'MyWorkspace', '-scheme', 'MySchemeA']
   source_directory: # for example, if this is necessary
   # any other per-module configuration
 - name: ModuleB
   build_tool_arguments:
    - -workspace
    - MyWorkSpace
    - -scheme
    - MySchemeB
```

Then there's some config code to write to look at these sources of modules, raise an error if the user has given multiple, and turn them into some single format for the rest of the code. 

(I wrote `custom_modules` as an example name here because I don't have any good ideas and it goes with `custom_categories` which is the other major config-file only thing...)

lib/jazzy/sourcekitten.rb
```diff
@@ -100,6 +110,33 @@ def self.group_type_categories(docs, type_category_prefix)
...
+        children, docs = docs.partition { |doc| doc.modulename == modulename }
+        make_group(
+          children,
+          modulename,
+          "",
+        )
+      end
+
+      # Get from the remaining docs if there are extensions that should also be part of this module.
+      group = group.compact.map { |group|
```
johnfairh, line 124: I get what this is doing; some issues/questions:

1. Using a child decl to identify the declaring module is neat, but not all extensions have children (think `extension SomeType : MyProtocol {}`).
2. When naming `newdoc` how about including the originating module name?  So it would be `MyModule+Swift.Int` or something -- otherwise not sure how the user knows where the extended type is from.  (Or even just `Swift.Int` given this is already nested under `MyModule`)
3. I don't understand the scenario on line 137 where there is stuff left over that doesn't belong to any of the modules.  What is an example of this?


About point 1, maybe think about changing how `parse()` works for multiple modules: could do each module's JSON one by one, then would always know which module is being worked on.  At the moment `group_docs_per_module` is really just *undoing* work that's already been done to combine everything.

lib/jazzy/config.rb
```diff
@@ -178,12 +178,9 @@ def hide_objc?
...
       default: []
 
-    config_attr :custom_modules,
-      # command_line: ['-cm', '--build-tool-arguments arg1,arg2,…argN', NArray],
-      description: 'Custom navigation categories to replace the standard ' \
-        "'Classes', 'Protocols', etc. Types not explicitly named " \
-        'in a custom category appear in generic groups at the ' \
-        'end.  Example: https://git.io/v4Bcp',
+    config_attr :modules,
+      description: 'Array of modules that are going to be documented ' \
```
IngoHaumann-TomTom, line 182: This description is unclear to me. I assume it's a list of modules, since array is an implementation concept, and they are not "being documented".
The line with the arguments puzzles me too, but maybe I need to dive into the tool more first.

lib/jazzy/config.rb
```diff
@@ -618,8 +611,9 @@ def validate
           '`framework_root` or `umbrella_header` may be ignored.'
       end
 
-      if custom_modules_configured && module_name_configured
-        raise 'Usage of config and comand line'
+      if modules_configured && module_name_configured
+        raise 'Jazzy only allows the use of a single command for generating documentation.' \
+        'Using both module configuration and modules configuration together is not supported.'
```
IngoHaumann-TomTom, line 616: This could also be a warning that we'll add the single module given to the list of modules and just proceed.

Issue #1368, opened 2023-10-02
Xcode 15: Unknown symbol kind swift.extension


esteluk:
With some code that offers an extension to a public (Foundation?) class, such as this:

```swift
public etension Bundle {
  static var myBundle: Bundle? {
    // ...
    return nil
  }
}
```

The following gets generated as in my symbol graph `build/symbolgraph/target/Target@Foundation.symbols.json`:

```json
{
    "kind": {
        "identifier": "swift.extension",
        "displayName": "Extension"
    },
    "identifier": {
        "precise": "s:e:s:So8NSBundleC29TargetE04selfD0ABSgvpZ",
        "interfaceLanguage": "swift"
    },
    "pathComponents": [
        "Bundle"
    ],
    "names": {
        "title": "Bundle",
        "navigator": [
            {
                "kind": "identifier",
                "spelling": "Bundle"
            }
        ],
        "subHeading": [
            {
                "kind": "keyword",
                "spelling": "extension"
            },
            {
                "kind": "text",
                "spelling": " "
            },
            {
                "kind": "typeIdentifier",
                "spelling": "Bundle",
                "preciseIdentifier": "c:objc(cs)NSBundle"
            }
        ]
    },
    "swiftExtension": {
        "extendedModule": "Foundation",
        "typeKind": "swift.class"
    },
    "declarationFragments": [
        {
            "kind": "keyword",
            "spelling": "extension"
        },
        {
            "kind": "text",
            "spelling": " "
        },
        {
            "kind": "typeIdentifier",
            "spelling": "Bundle",
            "preciseIdentifier": "c:objc(cs)NSBundle"
        }
    ],
    "accessLevel": "public",
    "location": {
        "uri": "...",
        "position": {
            "line": 6,
            "character": 7
        }
    }
}
```

Then when I run Jazzy with the config:
```yaml
swift_build_tool: symbolgraph
symbolgraph_directory: build/symbol-graph/target/
```

The following error occurs: 
```
/Users/nathanw/.rbenv/versions/3.2.2/lib/ruby/gems/3.2.0/gems/fastlane-2.213.0/fastlane_core/lib/fastlane_core/ui/interface.rb:153:in `shell_error!': [!] Exit status of command 'jazzy' was 1 instead of 0. (FastlaneCore::Interface::FastlaneShellError)
[...]

PR #1372, opened 2023-10-19
Symbolgraph extension symbols

johnfairh:
Support the opt-in symbolgraph extension symbols for Swift 5.9.

These are for extensions from other modules only, extensions in the same module don't get these symbols.  I've turned on the option for the test case because it runs on Swift 5.9 and exercises all the new paths -- looks good, grabs a bunch more information for those extensions including doc comments.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,7 +6,9 @@
 
 ##### Enhancements
 
-* None.
+* Support Swift 5.9 symbolgraph extension symbols.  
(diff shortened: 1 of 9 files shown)


johnfairh:
This all seems to work well.

Jazzy itself needs to support pre-5.9 swift so it does not set `-emit-extension-block-symbols` when asked to generate symbolgraphs itself.  The user can override this if necessary as done in the test runner - we could be smarter I suppose.


Fixes #1368

Issue #1373, opened 2023-11-21
zsh: no matches found: --documentation=./*.md


richardtop:
> jazzy --xcodebuild-arguments -scheme,MyCustomSchemeName --documentation=./*.md

> zsh: no matches found: --documentation=./*.md

What could be the issue here? I'm trying to avoid shell expansion when using this command:

> jazzy --documentation Full/Path_ToDoc.docc/**.md --xcodebuild-arguments -scheme,MyCustomSchemeName 

As this produces the following output:

>  Leftover unused command-line text: ["Full/Path_ToDoc.docc/File2.md", "Full/Path_ToDoc.docc/File1.md"]

Related: #1295, #850

richardtop:
Quoting solved the issue:

> jazzy --xcodebuild-arguments -scheme,MyScheme --documentation='Full/PathToDoc.docc/*.md'

PR #1374, opened 2023-11-21
Update README.md: add ticks to ensure that commands work in zsh

richardtop:
Related issue:
https://github.com/realm/jazzy/issues/1373

I've got confused because of the zsh errors and shell expansion. I've updated the examples in a way so that there is a better chance that they'll work for a beginner user.

### DIFF
--- a/README.md
+++ b/README.md
@@ -290,8 +290,8 @@ directory.
 
 | Description | Command |
 | ---         | ---     |
-| Command line option | `--documentation={file pattern}` |
-| Example             | `--documentation=Docs/*.md` |
+| Command line option | `--documentation='{file pattern}'` |
+| Example             | `--documentation='Docs/*.md'` |
 | jazzy.yaml example  | `documentation: Docs/*.md` |
 
 By default, jazzy looks for one of README.md, README.markdown, README.mdown or README (in that order) in the directory from where it runs to render the index page at the root of the docs output directory.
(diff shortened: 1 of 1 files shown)


johnfairh:
I don't think that's a very natural way to use quotes; they're not part of the program's syntax.

You could add a para to the 'troubleshooting' section for beginner shell users?  For zsh you could recommend the `alias jazzy='noglob jazzy'` trick.

richardtop:
Could you please explain "not part of the program's syntax" ?

`alias jazzy='noglob jazzy'` what would this do and how would it work? Where to put this code (zprofile?)

johnfairh:
The quote characters added are not part of jazzy's syntax, they're one of several techniques for avoiding unwanted shell expansion.

`noglob` turns off filename globbing in zsh. You could just type the command into a shell or I think `.zshrc` is the right place.  Lots of zsh documentation out there!  But it's not the place of a random tool's readme to tell users exactly how to use zsh or whatever shell they're using -- we just need a reminder that this kind of thing can interfere if you happen to be running Jazzy like this.

richardtop:
Clear. Unfortunately, this took a lot of time for me to figure out. Jazzy doesn't give a clear understanding in it's error descriptions, in fact, it just continues to generate the documentation. I think it's very important to have these issues fixed, as this is basically a default use case. ZSH is the default shell on mac and of course people don't want to have just a single readme file as an article.

johnfairh:
Doing in #1378.

PR #1385, opened 2024-05-04
Support Swift 5.10, move specs to Xcode 15.3

johnfairh:
* Update SourceKitten
* Update Ruby gems for murky cocoapods-related reasons
* Update CI & rebuild specs.  Changes addition of `any` in various places and something I don’t have time to understand: ‘json’ has started sometimes (but not always) escaping angle brackets in strings, so `<p>` comes out `\u003cp\u003e`.  This affects the `search.json` on-disk but not after deserialization.


johnfairh:
Danger is failing because I put this branch on the wrong repo, been too long.

PR #1394, opened 2024-06-22
Updates for Swift 6

johnfairh:
Code changes:
* SourceKit has broken SPI group extraction, have to fake it.
* SymGraph is generating duplicate conformances, have to uniq them.
* At some point `Actor` protocol changed to `AnyActor`
* SwiftPM moved the built module into a `Modules` directory…

Spec changes:
* Alamofire, RealmSwift - attribute ordering has been changed, in particular `@autoclosure` and `@escaping` have been reversed; more types are marked `@MainActor`.
* RealmObjC - many fewer symbols but these are all in “_private” header files eg. `RLMAppSubscriptionToken`, `RLMUpdatePolicy`.  I think Clang must have improved to respect an #ifdef or something.  Result is better.
* Moved jazzy test projects to require Swift 6

### DIFF
--- a/.github/workflows/Tests.yml
+++ b/.github/workflows/Tests.yml
@@ -34,6 +34,7 @@ jobs:
       - uses: actions/checkout@v4
         with:
           submodules: recursive
+          persist-credentials: false
       - uses: maxim-lobanov/setup-xcode@v1
         with:
           xcode-version: '16.0-beta'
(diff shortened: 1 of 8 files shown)


johnfairh:
Couple of weird inconsistencies vs. local; some will be arm vs. intel - leave it like this for a few betas...

PR #1401, opened 2024-09-19
[#1400] Group global constants and functions by file name

### LINKED ISSUE #1400: Group global constants and functions by the file they're in?
I see that specifying`custom_categories` can be used to achieve this, but for growing files, some that hold global constants and some that hold global functions, it is an additional maintenance overhead.

#1170 and #105 seem related, but having an option to group such symbols automatically by the file name / specifying a group at the file level would be amazing.

Is it possible to set up `custom_categories` using wildcards? i.e. maybe a pattern in the symbol names, or based on the file they're in?
If not, what would implementation for wildcards in `custom_categories` or file level grouping look like?

I believe since `custom_categories` is something that is already parsed and processed, adding support for wildcards or parsing files might be a smaller change?
Is it okay if I work on it?
Do we have access to the file name / source directory a symbol was in, in `def self.group_custom_categories(docs)`?
https://github.com/realm/jazzy/blob/4c0ee5dd0c67a566f025368b931547aab79f1cc9/lib/jazzy/sourcekitten.rb#L73-L88

Any pointers / tips would be appreciated.

eg:
```
"custom_categories": [{
        "name": "Global Utilities",
        "children": [
            "file:GlobalUtils.h",
            "regex:*CommonPattern*"
        ]
    }]
```

thePsguy:
- Make `doc_index` iterable.
- After `custom_categories` are processed, group any remaining global constants / functions further based on the file they're in.

Staged sample: https://serve-dot-zipline.appspot.com/asset/55d078ae-b243-597e-9a9d-257f931f2723/zpc/63fc25cv8le/GMSGeometryUtils%2BFunctions.html

Resolves #1400

### DIFF
--- a/lib/jazzy/grouper.rb
+++ b/lib/jazzy/grouper.rb
@@ -74,31 +74,34 @@ module Jazzy
       [group.compact, docs + globals_grouped_by_file]
     end
 
+    # Group root level Constants and Functions by the file they're declared in.
     def self.group_globals_by_file(docs, doc_index)
       types_to_files_to_symbols = {}
+      # Group symbols by file and type.
       doc_index.each do |index_item|
         if index_item.parent_in_code.nil? && (index_item.type.name == "Constant" || index_item.type.name == "Function")
           file_name = File.basename(index_item.file, ".*")
           types_to_files_to_symbols[index_item.type.plural_name] ||= {}
           types_to_files_to_symbols[index_item.type.plural_name][file_name] ||= []
-
           types_to_files_to_symbols[index_item.type.plural_name][file_name] << docs.delete(index_item)
-
         end
       end
 
+      # Create `SourceDeclaration`s for each file for each type of symbols it holds.
       types_to_file_groups = {}
       types_to_files_to_symbols.each do |type_name, files|
         types_to_file_groups[type_name] ||= []
         files.each do |file, declarations|
-          file_group = make_group(declarations, file, 'Globally available these types, are.')
+          file_group = make_group(declarations, file, '')
+          # Append the type this `SourceDeclaration` holds to the URL to avoid collisions when the same file holds both constants and functions.
           file_group.url_name = "#{file}+#{type_name}"
           types_to_file_groups[type_name] << file_group
         end
       end
 
+      # Create `SourceDeclaration`s for each type (Constants / Functions) to hold all the file level source declarations created above.
       type_groups = types_to_file_groups.map do |type_name, files_level_declarations|
-        make_group(files_level_declarations, type_name, 'These types are available globally.')
+        make_group(files_level_declarations, type_name, "The following #{type_name.downcase} are available globally.")
       end.compact
       
       type_groups
--- a/lib/jazzy/grouper.rb
+++ b/lib/jazzy/grouper.rb
@@ -69,7 +69,39 @@ module Jazzy
         children.each.with_index { |child, i| child.nav_order = i }
         make_group(children, category['name'], '')
       end
-      [group.compact, docs]
+
+      globals_grouped_by_file = self.group_globals_by_file(docs, doc_index)
+      [group.compact, docs + globals_grouped_by_file]
+    end
+
+    def self.group_globals_by_file(docs, doc_index)
+      types_to_files_to_symbols = {}
+      doc_index.each do |index_item|
+        if index_item.parent_in_code.nil? && (index_item.type.name == "Constant" || index_item.type.name == "Function")
+          file_name = File.basename(index_item.file, ".*")
+          types_to_files_to_symbols[index_item.type.plural_name] ||= {}
+          types_to_files_to_symbols[index_item.type.plural_name][file_name] ||= []
+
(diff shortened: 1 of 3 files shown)


lib/jazzy/grouper.rb
```diff
@@ -69,7 +69,42 @@ def self.group_custom_categories(docs, doc_index)
         children.each.with_index { |child, i| child.nav_order = i }
         make_group(children, category['name'], '')
       end
-      [group.compact, docs]
+
+      globals_grouped_by_file = self.group_globals_by_file(docs, doc_index)
```
thePsguy, line 73: Hi @johnfairh, should I place this under a new command line argument or something?

You recommended including a `glob/wildcard/regexp`, could you please elaborate a little more on what such a regexp would be used to filtered? Would it be used to determine whether constants in a file should be affected by this logic or not?

I can make that change in a following commit.

lib/jazzy/grouper.rb
```diff
@@ -69,7 +69,42 @@ def self.group_custom_categories(docs, doc_index)
...
+      [group.compact, docs + globals_grouped_by_file]
```
thePsguy, line 74: Adding `globals_grouped_by_file` to the end of docs makes the `Constants` and `Functions` sections show up at the end of other default sections, and also prevents the other default groups from being prepended with "Other".

The order of the sections, and file level groups in them appears to be consistent to be. Is there a specific way to ensure order is maintained that you would recommend?
I could sort the files in each group, and symbols in each file alphabetically to be sure. Should I?

cc: @johnfairh

lib/jazzy/grouper.rb
johnfairh: The 'glob' comment was in response to your example of `"children": [file:GlobalUtils.h"` which is not the feature you ended up going for.

The default behaviour mustn't change because of existing users, new behaviour can be behind a CLI option.

lib/jazzy/grouper.rb
```diff
@@ -69,7 +69,42 @@ def self.group_custom_categories(docs, doc_index)
...
+    end
+
+    # Group root level Constants and Functions by the file they're declared in.
+    def self.group_globals_by_file(docs, doc_index)
+      types_to_files_to_symbols = {}
+      # Group symbols by file and type.
+      doc_index.each do |index_item|
+        if index_item.parent_in_code.nil? && (index_item.type.name == "Constant" || index_item.type.name == "Function")
```
johnfairh, line 82: You can just use `docs` instead of `doc_index` if you just want top-level items, then won't need the `parent_in_code` check and will be more efficient.

lib/jazzy/grouper.rb
johnfairh: You need to go back to `group_docs()` and figure out the code there -- the advice to start in this routine was based (again :-)) on your other design.

Ruby hash iteration order is guaranteed - I think that is all you need to prove that you don't need any further sorting.

lib/jazzy/grouper.rb
```diff
@@ -69,7 +69,42 @@ def self.group_custom_categories(docs, doc_index)
...
+          types_to_files_to_symbols[index_item.type.plural_name][file_name] << docs.delete(index_item)
+        end
+      end
+
+      # Create `SourceDeclaration`s for each file for each type of symbols it holds.
+      types_to_file_groups = {}
+      types_to_files_to_symbols.each do |type_name, files|
+        types_to_file_groups[type_name] ||= []
+        files.each do |file, declarations|
+          file_group = make_group(declarations, file, '')
```
johnfairh, line 95: I _think_ this group is the one that should have the abstract "The following thingies are available globally" (you could write "are provided by <file>" given that you're adding source filenames into the UI.

And the group below (line 103) is the one that should have the abstract "The following files provide thingies" or something?

johnfairh:
This is a bit different to what I thought you were proposing in #1440 with your `"children": ["file:GlobalUtils.h", "regex:*CommonPattern*"]` example.  The big difference is that there, you were just adding stuff to existing groups.  Here you're trying to automatically fabricate a hierarchy of groups, and so have the additional problem of merging those into the overall group tree: it's a more ambitious problem!

The limitations of the left nav (only showing two levels) are not great here -- I'm not sure it's a step forward to not actually see (be able to search through) the list of constants (however long). (maybe if you can show a demo I'd change my mind here)

I feel that restricting the feature to just certain kinds of declaration (here, constants and functions) is not flexible enough for general use.  It would be OK to parameterise it I suppose, but I'm not sure why users would want to see a mixture of by-kind ("Other Classes/classes") and by-file ("Constants/Foo.swift/FooConstants").  

Options are (1) Group all left-over declarations by source-file instead of by kind; or (2) Group a configured set of kinds of left-over declarations by source-file, and group other kinds by kind.  

Both/either option needs a config option to modify the default behaviour.  Option (1) is a subset of (2), for (2) should probably have an 'all kinds' shortcut.

I don't understand what the demo is showing - the constants/functions pages just look empty to me?
Screenshot 2024-09-22 at 11 49 34

Code changes look in the right kind of direction - you clearly have a good grasp of how everything fits together here.

Issue #688, opened 2016-11-29
Does custom_categories support reg expression?


liuxuan30:
one question, it seems like custom_categories does not support reg expression?
Let's say I have many classes with some prefix: some starts with Line, some starts with Bars:

BarChartView.swift,
BarChartData.swift,
LineChartView.swift,
LineChartDataSet.swift

I want to group them into Line Type and Bar Type,
```
custom_categories:
  name: Bar Type
     children:
     - Bar*
  name: Line Type
     children:
     - Line*
```
It won't have any effect. Instead, I have manually list all the classes. Am I wrong?

istx25:
Hi @liuxuan30. Thanks for reaching out. I don't believe `custom_categories` supports regular expressions but I'm going to rope in @jpsim. I'm pretty new so he may have more to say.

liuxuan30:
If not supported yet, I guess it would be a good feature to have.. typing multiple names are killing me

jpsim:
We'd love a PR, @liuxuan30!

galli-leo:
I think this would make sense to add in combination with #624. Would it be ok, to include both in the same pr?

johnfairh:
Personally I'd prefer to see this as a separate PR, should be a simple change on top of the current code.

galli-leo:
@johnfairh Yeah I quickly tested it yesterday, theoretically it can be a one line change (if we just want to assume any item is a regex).

<sub>Sent with <a href="http://githawk.com">GitHawk</a></sub>

JUSTINMKAUFMAN:
@galli-leo @johnfairh Is there any movement on this? Would love to get it working.

PR #1418, opened 2025-08-28
Allow `custom_categories` to define a regex or a name for each category child

Enricoza:
Hi, I was seeing that #688 was still open and that a PR was welcome for that, so I've tried to have a go at it. 
Hopefully this can be merged as I think it could be a good addition and it would be very useful to me (and to others apparently).

Thank you guys for all your work on this!

---

A custom category can now specify a regex for each of its children by using an object with a regex key instead of just a string to find all of the declaration matching that regex.

For example:
```
custom_categories:
  - name: some_category_name
    children:
    - some_exact_name
    - regex: some_regex
```

Will create a category containing:
- the declaration "some_exact_name" (if it's found)
- all the declaration that match the "some_regex" regex. (e.g. "some_regex", "some_regex1", ...)


lib/jazzy/grouper.rb
```diff
@@ -49,22 +49,29 @@ def self.group_docs_by_module(docs, type_category_prefix)
...
-        children = category['children'].map do |name|
-          unless doc = doc_index.lookup(name)
-            warn 'WARNING: No documented top-level declarations match ' \
-              "name \"#{name}\" specified in categories file"
-            next nil
+        children = category['children'].map do |selector|
+          selected = if selector.is_a?(String)
+            unless doc = doc_index.lookup(selector)
+              warn 'WARNING: No documented top-level declarations match ' \
+                "name \"#{name}\" specified in categories file"
```
johnfairh, line 56: `name` is wrong here, cut + paste error - is causing the swift_spec test to fail

johnfairh:
Thanks for the PR, looks basically good!  I like the approach of using the yaml structure to do this.

You should be able to run `bundle exec rake rubocop` locally (or just `rake rubocop` depending on your setup) and work through the linter.

Please also add a change to the `README` to document this -- I realise the docs of `custom_categories` are 'minimal', not asking you to fix that, just to add a few lines explaining the addition.

(I expect to have some solid open source time in a few weeks and will be doing an Xcode26-ish release then, should be able to pull in this too.)

lib/jazzy/grouper.rb
```diff
@@ -49,29 +49,39 @@ def self.group_docs_by_module(docs, type_category_prefix)
...
     end
 
+    def self.select_docs(doc_index, selector)
+      if selector.is_a?(String)
+        unless single_doc = doc_index.lookup(selector)
+          warn 'WARNING: No documented top-level declarations match ' \
+            "name \"#{selector}\" specified in categories file"
+          []
+        end
+        [single_doc]
```
johnfairh, line 78: The `[]` on l76 has no effect, it's overwritten by l78 and the routine returns `[nil]` causing the crash downstream.

johnfairh:
Thanks this is great.  Don't worry about the _Danger_ failure, just one refactoring left to tidy up.

